PR #3, opened 2022-01-24
Update date to be 2022

NickFoden:
### Description:

- Date is showing as 2002
- Update date on index to be ` May 24-25, 2022`

Before:
image


joneidejohnsen:
Thanks for the PR, @NickFoden! We had to do some refactoring as well, so already fixed this in ff139a2eeae45e4311fe438c67ead231ce901ca4

PR #6, opened 2022-02-10
Dockerize dev setup

oyvind-stenhaug:
Adds a basic `docker-compose`-based setup that can be used for local development.

Testing this PR: See the new README.md file

### DIFF
--- a/studio/Dockerfile
+++ b/studio/Dockerfile
@@ -1,0 +1,11 @@
+FROM node:14
+
+RUN mkdir /app
+ADD . /app
+
+WORKDIR /app
+RUN yarn
+
+EXPOSE 3333
+
+CMD ["yarn", "sanity", "start", "--host=0.0.0.0"]
--- a/web/Dockerfile
+++ b/web/Dockerfile
@@ -1,0 +1,11 @@
+FROM node:14
+
+RUN mkdir /app
+ADD . /app
+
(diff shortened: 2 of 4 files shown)


studio/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM node:14
```
j0nas, line 1: Any reason we're not using the latest LTS? :slightly_smiling_face: (1/2)

```suggestion
# Node v16.14.0 LTS
FROM node@sha256:fd86131ddf8e0faa8ba7a3e49b6bf571745946e663e4065f3bff0a07204c1dde
```

web/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM node:14
```
j0nas, line 1: (2/2)
 
```suggestion
# Node v16.14.0 LTS
FROM node@sha256:fd86131ddf8e0faa8ba7a3e49b6bf571745946e663e4065f3bff0a07204c1dde
```

j0nas (approved):
Thank you for submitting this! One small comment, otherwise approved.

studio/Dockerfile
j0nas: I figured there'd be a reason! Ok, makes good sense to me. :slightly_smiling_face:

PR #10, opened 2022-02-14
Enable PostCSS processing of custom CSS media and properties

oyvind-stenhaug:
##Summary

This PR intends to enable PostCSS processing of `@custom-media` and CSS custom properties (defined on `:root`).

Could optionally drop the latter, since we have confirmation that IE11 can be ignored.

## Testing this PR

1. `docker-compose up --build web`
2. Add a CSS rule somewhere that uses a custom media query and/or custom property. E.g. in web/components/Heading/Heading.module.css, add the line `@media (--large-up) { .heading { color: var(--color-brand-orange) } }`
3. `docker-compose exec web bash` in a separate terminal window, and `yarn run build`
4. Verify that the rule from step 2 was transformed so as to no longer include any `--` parts, e.g. `grep -or 'Heading_heading.[^}]*}' .next/static/css/` and `grep -or '@media.[^)]*)' .next/static/css/`

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -27,6 +27,8 @@
     "eslint": "8.6.0",
     "eslint-config-next": "12.0.8",
     "postcss": "^8.4.5",
+    "postcss-custom-media": "^8.0.0",
+    "postcss-custom-properties": "^12.1.4",
     "prettier": "^2.5.1",
     "tailwindcss": "^3.0.14",
     "typescript": "^4.5.5"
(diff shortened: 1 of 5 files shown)


j0nas (approved):
Great job adding this! :100: I used your example media query for `Heading`, and the `grep` output shows only `{color:#ffba00}` and no sign of the original variable.

> Could optionally drop the latter, since we have confirmation that IE11 can be ignored.

I see to reason to drop it when it's this simple to add it and the work's already been done. :slightly_smiling_face:

PR #12, opened 2022-02-15
Links to use entities' `slug`s, not `_id`s

j0nas:
# Links to use entities' `slug`s, not `_id`s

## Intent

@kmelve added `slug` fields for the main entities in Studio. I've generated slugs for all People, Sessions and Venues in Studio, and this PR refactors the codebase to consistently use these slugs rather than the previously used `_id` or `title` fields.

## Description

This was mainly achieved using the added `getEntityPath` function. Centralizing URL generation/handling has worked out well for me in the past, and lack thereof has proven painful on projects in the long run.

This PR additionally: 
- renames the type `Speaker` -> `Person` to align with Studio terminology
- Simplifies some GROQ queries and more consistently uses existing _types/_  rather than re-declaring in `interface`s
- Deletes the unused _/web/pageResources/program/Venues/_ directory
- Refactors _/web/components/TextBlock/TextBlock.tsx_ 

## Testing this PR

1. `docker-compose up`
2. Go to [http://localhost:3000/home](http://localhost:3000/home)
3. Click on various (entity, where possible) links and verify that you can navigate correctly between entities and that you aren't linked to any 404'ing pages
4. Open [/sessions/unifying-people-with-structured-content](http://localhost:3000/sessions/unifying-people-with-structured-content) and verify that the contents, especially the session description correctly renders in accordance to what's stored in the "Long description" field of [the corresponding Session in Studio](http://localhost:3333/desk/session;a810fe51-acbc-4bb6-a213-a94ad9b5c2b2)
    * Note that this PR is not supposed to account for all possible content that this field could contain, content support will be handled by a different PR; e5374c099848d8d8c11070585a289e4a362f2d79 refactors `<TextBlock />` because it was discovered to have broken during the refactoring process in this PR


web/pages/speakers/[slug].tsx
```diff
@@ -60,8 +53,10 @@ const Speakers = ({
             <Heading>{name}</Heading>
             <div>{title}</div>
             <div>
-              <Link href={`/speakers/${name}`}>
-                <a>twitter.com/{twitter}</a>
+              <Link href={`https://twitter.com/${twitter}`}>
```
Comment (oyvind-stenhaug) [inline web/pages/speakers/[slug].tsx:56]: Is there any reason to use `Link` for an external URL? Or could `href` instead be set on the `<a>` below?

web/pages/speakers/[slug].tsx
```diff
@@ -5,27 +5,19 @@ import SectionBlock from '../../components/SectionBlock';
...
 const QUERY = `
   {
-    "speaker": *[_id == $_id][0] {
-      _id,
-      name,
-      title,
-      bio,
-      "twitter": social.twitter,
-      photo,
+    "speaker": *[slug.current == $slug][0] {
```
Comment (oyvind-stenhaug) [inline web/pages/speakers/[slug].tsx:17]: Sanity slug uniqueness is based on document type, so I think this query should specify `_type` as well. Something like this (not tested):

```suggestion
    "speaker": *[_type == 'person' && slug.current == $slug][0] {
```

oyvind-stenhaug (changes requested):

web/pages/speakers/[slug].tsx
Comment (j0nas) [reply web/pages/speakers/[slug].tsx]: No, as far as I gather there's no reason to use `Link` for external links. _speakers/index.tsx_ already uses an `a`, thanks for catching this inconsistency.

oyvind-stenhaug (approved):

Issue #7, opened 2022-02-14
Warning about mismatching timestamps on front page


oyvind-stenhaug:
Opening http://localhost:3000/home locally yields the following in browser console:

```
Warning: Text content did not match. Server: "May 25, 2022 12:08" Client: "May 25, 2022 02:08"
```

Question here is really which time zone should be used when displaying the start/end date. If it's the user's time zone (and here I mean "time zone" as in e.g. "Europe/Oslo"), then how does that work with Next.js static rendering?

oyvind-stenhaug:
For now, at least, it's showing UTC days, and warning is fixed. By https://github.com/sanity-io/structured-content-2022/pull/17

PR #13, opened 2022-02-15
Frontpage header

oyvind-stenhaug:
# Frontpage header

## Intent

Implement the page-specific header of the frontpage, from the first iteration of the design.

## Description

Splits the first SectionBlock of the front page out as a separate component, and implements HTML and CSS for that component based on the current design sketches in Figma. Behavior for narrower screens is simplified, no finished sketches for that quite yet.

Although the name "ConferenceHeader" has "Header" in it, I don't expect it to be common across all pages. Thus I chose to move it out of the `<header>` element, which corresponds to the `banner` ARIA role.

Not sure about the placement of this component in the file structure. I see there are some things called "pageResources". Are they intended to be components that are currently only used on one page, and then if you want to reuse the component somewhere else you have to move it into `web/components`? Or is it something else?

I'm thinking that if the front page is changed into a "landing page" in Knut's regime, then the "page-specific" concept might no longer apply. This element may then end up corresponding to some "Section" instead… but this still seems unclear.

I'm trying to focus on the frontend parts here, and not so much on the relation between the actual underlying content and the structure in Sanity. Specifically, the start/end date contents do not correspond to the Figma sketch; I left the old format helper function intact. See e.g. https://github.com/sanity-io/structured-content-2022/issues/7

## Testing

1. `docker-compose up --build web`
2. Open http://localhost:3000/home
3. Verify that the section between "Home" etc and "Sign up for Structured Content 2022" looks OK


web/components/ConferenceHeader/ConferenceHeader.tsx
```diff
@@ -0,0 +1,44 @@
...
+  name,
+  startDate,
+  endDate,
+  description,
+}: ConferenceHeaderProps) => {
+  return (
+    <div className={styles.root}>
+      <h1>
+        {/* eslint-disable-next-line @next/next/no-img-element */}
+        <img
```
j0nas, line 22: Two questions, mainly out of curiosity:
1. Why don't we use `next/image` here?
2. Since this content is very much "above the fold", should we consider inlining this SVG? (Inlining as a base64-encoded string in the HTML, that is. Not the "componentization" done in www-sanity-io which led to SVGs being parsed as JS)

j0nas (approved):
Looks great! One optional remark.

web/components/ConferenceHeader/ConferenceHeader.tsx
oyvind-stenhaug: 1. In short, because it's an SVG.

    In long: To me, next/image adds quite a bit of complexity when it comes to styling. E.g. it seems to be unable to size an image based on the height, so you have to hack around it by setting width to desired height * aspect ratio. There was a clipping bug on sanity.io related to this.
    
    Looking at its [main selling points](https://nextjs.org/docs/basic-features/image-optimization), they seem to be
    - "Always serve correctly sized image for each device": irrelevant for SVG as long as it's not transformed in some way to lose detail on smaller viewports. Here, it seems to set `srcset` and serve the exact same image on a different URL, which is pointless at best and wasteful at worst
    - "Prevent Cumulative Layout Shift automatically": we can just set the `width` and `height` attributes, that's a simpler way of accomplishing the same thing nowadays
    - "Images are only loaded when they enter the viewport": irrelevant here. Might still be useful in general, though it would be simpler to use `loading="lazy"` once Safari < 15.4 usage is sufficiently low
    - "with optional blur-up placeholders": irrelevant for SVG. Here, it seems to add an extra `` with a completely blank image that seems like a waste to decode and render
    - "On-demand image resizing": same as first bullet point

2. Perhaps, but it's not clear to me if it's necessarily better or how much difference it would make. Filed a GitHub issue on it. (Using a base64-encoding sounds unnecessary, though; in this particular case it looks like it would add about 35% to the size. But it's possible to have `data:` URIs without base64.)

j0nas: Great, thanks for the in-depth explanation!

PR #16, opened 2022-02-16
Content model iteration

kmelve:
- Harmonized some field naming
- Added more fields to the content model
- Content Spec Sheets WIP
- Tweaks for desk structure

### DIFF
--- a/studio/schemas/documents/event.ts
+++ b/studio/schemas/documents/event.ts
@@ -1,7 +1,10 @@
+import { CalendarIcon } from "@sanity/icons";
+
 export default {
   name: "event",
   title: "Events",
   type: "document",
+  icon: CalendarIcon,
   groups: [
     {
       name: "messaging",
@@ -98,7 +101,7 @@ export default {
       title: "End date",
       group: "practical",
     },
-    {
+    /* {
       name: "microcopy",
       type: "array",
       title: "Microcopy",
@@ -174,6 +177,6 @@ export default {
           ],
         },
       ],
-    },
+    }, */
   ],
 };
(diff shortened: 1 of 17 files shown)


studio/schemas/documents/event.ts
```diff
@@ -98,7 +101,7 @@ export default {
       title: "End date",
       group: "practical",
     },
-    {
+    /* {
```
j0nas, line 104: This code should probably be deleted instead of being commented out.

j0nas (commented):
This looks great! :100: Also love the Preview functionality that's been set up.

I don't really understand the "Content Specification Sheet" bit, wish maybe the PR description went into a bit more details. But that bit might not be relevant for frontend (yet)?

Couple things I noticed: 
- Could the various fields that no longer are in use, be removed from the relevant documents? E.g.: 
image
- _People_ only show the Preview; no _Editor_ tab is available in the Studio
- _Sessions_ and _Venues_ don't have a _Preview_ tab
- Content: a few documents have no "Internal name" values, making them all list as "Untitled" -- could these get named?
- There is a "Editorial Articles" category -- is/will this be used for anything, or can it be removed?
- Can Sponsorships get a `sortIndex` field or something similar, so that frontend doesn't have to hardcode/"understand" which sponsorship tier is better and ad-hoc sort them? 
- Can the entities that have a `slug` field have this field be made required?


As always, these changes will be easier to reason about when (more) Sanity data is added. If you can, please prioritise that whenever possible.

I don't have many comments on this PR's code itself :slightly_smiling_face:

PR #18, opened 2022-02-16
Header nav

oyvind-stenhaug:
# Header nav

## Intent

Implement frontend for the "global" header, i.e. what's currently known as the `Nav` component in the code. Not including the "sticky" part.

## Description

This is "Desktop/Header", "Tablet/Header" and "Mobile/Menu trigger" + "Mobile/Menu" in the Figma sketches.

Known issues:
- As discussed on Slack, the body is still scrollable when the mobile menu is open, which does not make for a good experience. This indicates that the menu open/close state should really be lifted higher up, so that body `overflow` can be toggled accordingly. Given the way `<Nav>` is currently included repeatedly per page rather than in `_app.tsx`, it's not clear to me how this should best be structured. Particularly because there's also a wish for the header to be initially invisible on the front page and only be toggled on once the user has scrolled down past a certain point.
- The three links starting with "Registration info" in the mobile menu Figma sketch are not included here. These are intentionally left out for now, per [Slack thread from today](https://wja.slack.com/archives/C02E2ERKR7E/p1645005357175499)

## Testing this PR
1. `docker-compose up --build web` (based on what we discussed on Slack, it's possible you won't need the `--build`, but I do; otherwise the CSS variables are not re-validated)
2. Open http://localhost:3000/home and verify that the topmost header/menu looks and behaves OK both with a wide window and with a narrow one (aside from known issues)


web/components/ButtonLink/ButtonLink.tsx
```diff
@@ -0,0 +1,13 @@
+import Link from 'next/link';
+import styles from './ButtonLink.module.css';
+
+interface ButtonLinkProps {
+  text: string;
+  url: string;
+}
+
+export const ButtonLink = ({ text, url }) => (
```
j0nas, line 9:
```suggestion
export const ButtonLink = ({ text, url }: ButtonLinkProps) => (
```

web/components/Nav/Nav.module.css
```diff
@@ -1,22 +1,138 @@
...
+
+.link:hover,
+.link:focus {
+  text-decoration: underline;
 }
 
-@media screen and (max-width: 480px) {
-  .nav__items {
-    max-width: 200px;
+@media not all and (--medium-up) {
```
j0nas, line 42: Nit: this reads very strange to me. Might be because I'm not familiar with the spec ([I just read up on this](https://developer.mozilla.org/en-US/docs/Web/CSS/Media_Queries/Using_media_queries#inverting_a_querys_meaning)), but can we create a `--medium-down` variable instead, so that we can simply write `@media (--medium-down)`? (If medium-down is what we want, that is.)

web/components/Nav/Nav.tsx
```diff
@@ -1,27 +1,79 @@
...
-    </ul>
-  </nav>
-);
+export const Nav = () => {
+  const [menuOpened, setMenuOpened] = useState(false);
+  const contentsId = 'nav-menu-contents';
+
+  return (
+    <nav className={styles.nav}>
+      <PageContainer>
```
j0nas, line 15: Why is `PageContainer` used here?

j0nas (changes requested):
Cool! :100: 

One small thing I've noticed is that the sketches place the "mobile" Menu button in the footer, rather than in/above the header as this PR does. I might absolutely be wrong here, but I'd guess that this would go for all pages:
image


And the `->` is not rendered with an arrow ligature in the md+ "Tickets" button: 
image


In the "Website (Sketches)" page in Figma, in this latest version of the /about page, the header seems to take up the full width of the viewport: 
image

This PR has it contained to being within `PageContainer`. In #19, I wrap each page in `<></>` and place the `<Footer />` component outside the `PageContainer`.

oyvind-stenhaug:
> One small thing I've noticed is that the sketches place the "mobile" Menu button in the footer, rather than in/above the header as this PR does.

Oh, very good point, thanks for catching this! I didn't notice the copy in the footer, not sure if I looked before it was there or if I just missed it.

This could be fixed/implemented in a separate PR, maybe? After the _document/_app refactoring we discussed.

> And the `->` is not rendered with an arrow ligature in the md+ "Tickets" button: image

That's strange. Oh, it sems like a browser-specific thing, it works for me in Firefox but I assume you used Chrome or something like that. Looks like it's the custom letter spacing that breaks it, so I'll remove that here and test it in a few browsers. Worst case, it'll have to be an icon.

> In the "Website (Sketches)" page in Figma, in this latest version of the /about page, the header seems to take up the full width of the viewport: image
> 
> This PR has it contained to being within `PageContainer`. In #19, I wrap each page in `<></>` and place the `<Footer />` component outside the `PageContainer`.

Yeah. I only updated `home` here, I've focused on that page specifically. Partially to avoid causing merge conflicts. I can copy it to every page, though.

web/components/Nav/Nav.tsx
j0nas: Rename!

oyvind-stenhaug:
- Ticket button fixed in https://github.com/sanity-io/structured-content-2022/pull/18/commits/1e56012519c5c8bea3bf47d716b8252e7f7abc6f
- Header fixed for non-home pages in https://github.com/sanity-io/structured-content-2022/pull/18/commits/23b98bdcb6f5a940560f0dae420d5e6d7160d6f4

web/components/GridWrapper/GridWrapper.tsx
```diff
@@ -0,0 +1,10 @@
+import { ReactNode } from 'react';
+import styles from './GridWrapper.module.css';
+
+interface GridWrapperProps {
+  children: ReactNode;
+}
+
+export const GridWrapper = ({ children }) => (
```
j0nas, line 8:
```suggestion
export const GridWrapper = ({ children }: GridWrapperProps) => (
```

j0nas (approved):
LGTM!

web/components/GridWrapper/GridWrapper.tsx
oyvind-stenhaug: Thanks… too bad lint doesn't complain about this :)

PR #19, opened 2022-02-16
Implement Footer component in accordance with Figma sketches

j0nas:
# Implement Footer component in accordance with Figma sketches

## Intent

Implements the Footer as described in [Figma](https://www.figma.com/file/XjAvJKsLpxYJSc8QDWK81j/Sanity-SCC-Website-(Shared)?node-id=0%3A1). 
image

## Description

In the interest of keeping this PR smaller, two things have been postponed to a future PR:
- A restructuring of the pages, putting the footer into `<PageContainer />` or even <em>_document.tsx</em> or <em>_app.tsx</em>. This isn't very difficult to do, but would perhaps clutter this PR a bit
- Adding a "sticky" effect to the footer for cases where the content doesn't fill the entire page; e.g. for current http://localhost:3000/test -- this could be handled by setting `min-height: calc(100vh - <FOOTER_HEIGHT>px);` on the content wrapper. Because this point dovetails so nicely with the point above, handling this case has been omitted from this PR

## Testing this PR

1. (If using docker-compose, `docker-compose build web`)
2. Start the application (`docker-compose up`)
3. Verify that each page that you can navigate to renders the new footer
4. Verify that the new footer is in accordance with the design sketches in every reasonable aspect for every "defined" viewport/breakpoint


j0nas (commented):
Your changes look great, @oyvind-stenhaug. Merge at will. :slightly_smiling_face:

PR #20, opened 2022-02-17
Move <header> to App

oyvind-stenhaug:
# Move <header> to App

## Intent

Moves the shared "global" `<header>`, along with the `<Nav>` inside, to the `<App>` component. This is in preparation for implementing the sticky header (and/or footer placement), to try to avoid duplicating any logic/styling that may need to exist outside.

## Description

Earlier we discussed moving things into `_document`, but it looks like that's not going to work very well according to [the docs](https://nextjs.org/docs/advanced-features/custom-document#caveats). I accidentally committed/pushed a Tailwind class removal in that file regardless, but might as well leave that in.

Having the shared stuff in App seems to be the intended way according to the aforementioned docs. Previously we didn't really know if the header/footer would be shared, but now we do. And hopefully this new placement will enable adjusting the overall page layout (e.g. introducing extra wrappers, styling or logic if necessary) without touching all the `web/pages`.

This change _does_ touch all the web/pages, though, hence scoping this PR to try to get that part over with as soon as possible.

## Testing this PR

1. `docker-compose up web`
2. Open http://localhost:3000/ and a few different pages, e.g. clicking the links in the header menu, and verify that there is exactly one such header menu on each page

### DIFF
--- a/web/pages/_app.tsx
+++ b/web/pages/_app.tsx
@@ -2,9 +2,10 @@ import { useEffect } from 'react';
 import { getCookieConsentValue } from 'react-cookie-consent';
 import TagManager from 'react-gtm-module';
 import CookieConsent from '../components/CookieConsent';
+import Nav from '../components/Nav';
 import '../styles/globals.css';
 
-function MyApp({ Component, pageProps }) {
+function MyApp({ Component, pageProps, router }) {
   useEffect(() => {
     const hasConsent = getCookieConsentValue();
     if (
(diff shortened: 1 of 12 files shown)


web/pages/_app.tsx
```diff
@@ -2,9 +2,10 @@ import { useEffect } from 'react';
 import { getCookieConsentValue } from 'react-cookie-consent';
 import TagManager from 'react-gtm-module';
 import CookieConsent from '../components/CookieConsent';
+import Nav from '../components/Nav';
 import '../styles/globals.css';
 
-function MyApp({ Component, pageProps }) {
+function MyApp({ Component, pageProps, router }) {
```
j0nas, line 8:
```suggestion
function MyApp({ Component, pageProps }) {
```

j0nas (approved):
LGTM!

web/pages/_app.tsx
oyvind-stenhaug: Ah yes, thanks… I accidentally committed some experimental code. Missed this part.

PR #21, opened 2022-02-17
Sponsor image sizes

j0nas:
# Sponsor image sizes

## Intent

Implement the correct sizes for the various Sponsorship types based off the Figma sketches.

## Description

I struggled a bit with getting `imageUrlFor` to provide the desired size until I noticed that a `rect=` parameter was being included in the generated URL. This parameter crops the image in weird, undesirable ways. I used a regexp in `<Sponsor />` to remove it, which finally fixed the issues I was experiencing.

@oyvind-stenhaug Although the intention was to only fix the image dimensions, I found it easiest to verify the correctness of the implementation by adhering fully to the Figma sketches. If you want, you can disregard any CSS/markup issues and fix them yourself, by either directly committing to this PR, or later in a follow-up PR.


web/components/Sponsor/Sponsor.tsx
```diff
@@ -1,22 +1,56 @@
...
+  const src = imageUrlFor(image)
     .auto('format')
-    .height(100)
-    .fit('clip')
-    .url();
+    .bg('fff')
+    .size(dimension.width, dimension.height)
+    .fit('max')
+    .url()
+    .replace(/rect=\d+,\d+,\d+,\d+&/, '');
```
oyvind-stenhaug, line 48: Hm, come to think of it, I think I remember seeing someone on Sanity's Slack also having issues with the image URL builder… Would using [ignoreImageParams](https://www.sanity.io/docs/image-url#ignoreimageparams) work instead, as a slightly cleaner way of getting rid of the rect parameter?

oyvind-stenhaug (commented):
Thanks! I can probably take over the branch to clean this up frontend-wise, as you suggest. Just one minor comment about the image URL thing.

web/components/Sponsor/Sponsor.tsx
j0nas: Great suggestion, I didn't know of it. b1a87c1

oyvind-stenhaug: Yeah I wouldn't either if not for that Slack thread.

j0nas:
Changes look good, @oyvind-stenhaug. 

(We might want to file an issue on the width/height-thing to fix later, and perhaps also an issue about migrating to using `<h1-6>`s vs. the `Heading` component.)

PR #22, opened 2022-02-17
Sticky header

oyvind-stenhaug:
# Sticky header

## Intent

Make the header "sticky" as specified by the designer, and make a couple other small fixes to the header/menu

## Description
- Disables scrolling the viewport when the mobile menu is open, to avoid wonky behavior there (e.g. when swiping on my phone I would get a scrollbar and the menu jumped around)
- Fixes a bug where the mobile menu didn't close after a link was clicked
- Moves mobile menu button to the bottom of the viewport, instead of the top right of the page. This button is fixed at that position on all pages.
- On most pages, the header/menu for wider screens is likewise fixed to the viewport, only to the top instead.
- On the front page in particular, however, the header/menu is completely hidden unless the page is scrolled sufficiently far down. (Round about the point where the logo is scrolled out of the screen.)

I enabled this last exception both for `/home` and for `/` which will be the new home page. Upside of this is that we don't have to remember it once /home is retired, but downside is that there is no menu until the page is tall enough… (except mobile)

## Testing this PR

1. `docker-compose up web`
2. Open http://localhost:3000/home and verify that the header is absent
3. Scroll down and verify that the header eventually appears
4. Click the link to another page (e.g. Program) and verify that the header is present at all times
5. Shrink window to mobile-ish width and verify that the menu button appears at the bottom (if you've clicked away the cookie consent dialog)
6. Click the Menu button and verify that the viewport is no longer (vertically) scrollable, i.e. that you can't scroll the page that still exists behind the menu
7. Click a link and verify that the menu is closed, navigation to the linked page happens, and viewport scrollbar is back if the page is tall enough to require it (e.g. About should work, unless the screen is extremely tall!)

### DIFF
--- a/web/components/Nav/Nav.module.css
+++ b/web/components/Nav/Nav.module.css
@@ -1,4 +1,5 @@
 .nav {
+  background: var(--color-brand-white);
   box-shadow: 0 6px 0 rgba(0, 0, 0, 0.1);
 }
 
@@ -41,13 +42,14 @@
 
 @media not all and (--medium-up) {
   .nav {
+    background: none;
     box-shadow: none;
   }
 
   .menuButtonWrapper {
     display: block;
-    text-align: right;
-    padding: 24px 0;
+    padding-bottom: 46px;
+    text-align: center;
   }
 
   .menuButton,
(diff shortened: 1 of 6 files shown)


j0nas (approved):
LGTM! 

Only slightly weird thing I noticed was that on /home in large viewport (in Chrome), if you scroll down and refresh (so that the page loads scrolled down), the sticky menu would not appear, and it only appears if you scroll back up and down again. Not sure if this is something worth spending time on to fix at this point.

oyvind-stenhaug:
Hm, odd. Yeah, I've filed a GitHub issue on it instead…

PR #27, opened 2022-02-18
NavBlock frontend

oyvind-stenhaug:
# NavBlock frontend

## Intent

Adds frontend code for the "nav block", i.e. the grid of yellow boxes near the top of the front page in the Figma sketches

## Description

This is not the prettiest markup, unfortunately. There is no underlying system or logic to how the boxes are wrapped for different screen sizes (the design seems quite reliant on the specific link texts to be suc h as they are in Figma, w.r.t. text length, amount etc). So this is all hard-coded.

In short, I use a UL to try to be sensible w.r.t. semantics/accessibility. But then I have to add extra empty list items because of all the empty boxes, because it's not really possible to create such boxes in CSS alone (`:before` and `:after` is limited to one per element, and they are contained within the element that the base selector matches, they're not siblings, so it doesn't work here). To try to limit the "damage" these are labelled with `aria-hidden`.

To control which boxes go in which lines (without losing the semantic UL), I had to use extra list items to act as "separators" too. ([Relevant blog post](https://tobiasahlin.com/blog/flexbox-break-to-new-row/)). These are 100% width, and then the rest of the boxes flex as required to fill the line they're on (actual links only flex down, filler boxes also flex up).

Initially I tried using `:nth-child` (along with labelling each `li` with a comment), but I think the `mobile`/`tablet`/`desktop` classes are probably a bit less hard to read. (Basically these are used to specify when a given "fake"/separator element is supposed to apply.)

There are a few differences between the links here and the ButtonLink, so it seemed like a toss-up between just implementing these links independently and trying to introduce a special-case variant of ButtonLink to import. I went for the former here. (I think the identical bits are basically colors and text-decoration. Other than that, all the details are different.)

## Testing this PR

1. Start the application locally, e.g. `docker-compose up web`
2. Open [localhost:3000/home](http://localhost:3000/home) in your browser
3. Verify that the yellow link grid ("Program" etc) looks OK in different screen sizes and that the links work as expected

(It's of course also possible to test it on https://structured-content-2022-web-git-nav-block-frontend.sanity.build/home)

### DIFF
--- a/web/components/NavBlock/NavBlock.tsx
+++ b/web/components/NavBlock/NavBlock.tsx
@@ -18,15 +18,15 @@ export const NavBlock = () => (
           styles.desktop
         )}
         aria-hidden="true"
-      ></li>
+      />
       <li
         className={clsx(styles.fakeItem, styles.tablet, styles.desktop)}
         aria-hidden="true"
-      ></li>
+      />
       <li
         className={clsx(styles.fakeItem, styles.desktop)}
         aria-hidden="true"
-      ></li>
+      />
       <li
         className={clsx(
           styles.divider,
@@ -35,7 +35,7 @@ export const NavBlock = () => (
           styles.desktop
         )}
         aria-hidden="true"
-      ></li>
+      />
       <li
         className={clsx(
           styles.fakeItem,
@@ -44,15 +44,15 @@ export const NavBlock = () => (
           styles.desktop
         )}
         aria-hidden="true"
-      ></li>
+      />
       <li
         className={clsx(styles.fakeItem, styles.tablet, styles.desktop)}
         aria-hidden="true"
-      ></li>
+      />
       <li
         className={clsx(styles.fakeItem, styles.desktop)}
         aria-hidden="true"
-      ></li>
+      />
       <li className={styles.item}>
         <Link href="/speakers">
           <a className={styles.link}>Speakers</a>
@@ -66,7 +66,7 @@ export const NavBlock = () => (
           styles.desktop
         )}
         aria-hidden="true"
-      ></li>
+      />
       <li className={styles.item}>
         <Link href="/about">
           <a className={styles.link}>About</a>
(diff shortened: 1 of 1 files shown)


web/components/NavBlock/NavBlock.tsx
```diff
@@ -0,0 +1,103 @@
+import Link from 'next/link';
```
j0nas, line 1: Could this file could be made shorter if you extract the various repeated elements into their own components, either directly in this file or in a subdirectory. With the former perhaps requiring less effort for sharing styles.

Example of what I'm imagining
``` 
const FakeItem = ({ mobile }: { mobile?: boolean }) =>
  <li
    className={clsx(
      styles.fakeItem,
      mobile && styles.mobile,
      styles.tablet,
      styles.desktop
    )}
    aria-hidden="true"
  />;
``` 

And then usage could simply be `<FakeItem mobile />` etc

(This is a first-glance suggestion, YMMV. I might have misjudged how many SLOC this will save.)

web/components/NavBlock/NavBlock.tsx
```diff
@@ -0,0 +1,103 @@
...
+          styles.mobile,
+          styles.tablet,
+          styles.desktop
+        )}
+        aria-hidden="true"
+      ></li>
+      <li
+        className={clsx(styles.fakeItem, styles.tablet, styles.desktop)}
+        aria-hidden="true"
+      ></li>
```
j0nas, line 25: Nit: In JSX, empty XML nodes should be collapsed.<sup>[citation needed]</sup> 
My editor gets a little upset otherwise. But this is personal (jsx) preference.

j0nas (approved):
Impressive. This looked really difficult to implement, and you seem to have done it well. It is behaving exactly as expected and in accordance with what the sketches specify.

Not sure how Behalf imagined this behaving with dynamic content (as [my discovery](https://wja.slack.com/archives/C02AX66E2S0/p1645186109237529) seems to imply is eventually expected), but that's a challenge for future iterations. :slightly_smiling_face: 

A couple simple comments on _NavBlock.tsx_, which you may choose to

web/components/NavBlock/NavBlock.tsx
oyvind-stenhaug: Hm, I think that sounds like a really good idea. (And probably with a preference for having in the same file, yes.) And similarly for `<Divider … />`

I suspect you'd be faster at this than me, so I'll appreciate it if you can help out if there's time. For now I'll merge it as-is.

web/components/NavBlock/NavBlock.tsx
oyvind-stenhaug: Thanks, JSX is kind of an "uncanny valley" :) 249a1f591e29f1d85a260039031ca0cdf7540a46

PR #36, opened 2022-02-21
ConferenceUpdatesForm frontend

oyvind-stenhaug:
# ConferenceUpdatesForm frontend

## Intent

Update markup and CSS for ConferenceUpdatesForm so it more-or-less looks like the Figma sketch

## Description

Not much to say here. Markup is mostly unchanged aside from the introduction of the `GridWrapper` and some ARIA labels.

Would be a bit nicer to keep the original button label and only add the arrow via CSS, but the aria-label was faster.

## Testing this PR
1. Open [/home](https://structured-content-2022-web-git-conference-updates-form-92b5da.sanity.build/home)
2. Scroll down almost to the bottom
3. Verify that the "Get conference updates" section above the gray footer looks OK in various screen sizes

### DIFF
--- a/web/components/ConferenceUpdatesForm/ConferenceUpdatesForm.module.css
+++ b/web/components/ConferenceUpdatesForm/ConferenceUpdatesForm.module.css
@@ -1,22 +1,94 @@
 .container {
-  display: flex;
-  align-items: center;
-  justify-content: center;
+  text-align: center;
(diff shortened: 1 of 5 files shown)


j0nas (approved):
Looks great. :slightly_smiling_face: Cool to see pieces of the puzzle landing in place.

PR #46, opened 2022-02-23
Fetch & list Sanity Program details, delete content breaking CI

j0nas:
# Fetch & list Sanity Program details, delete content breaking CI

## Intent

- Deletes /speakers page that seems to be unused both in approved and "In Review" Figma sketches. Fixes CI after d9682ed4f41c72bb8094821ee4cc04c08ce91d3c.
- Deletes all contents on ad-hoc /program page, replacing it with an attempt at copying the accordion component under "Program Schedules" in the Figma sketch

## Details

There are still some things not clear to me (e.g. Sessions having `duration` fields in Sanity vs. from->to times expected by Figma sketches) but I suspect that is due to the modeling process [still being WIP](https://wja.slack.com/archives/C02UKELNTE0/p1645572982557739). 

## Testing this PR

1. Open [/program](https://structured-content-2022-web-37shabol3.sanity.build/program)
2. Verify that an accordion with some content matching the [Programs](https://structured-content-2022-studio-n4hn2lwwa.sanity.build/desk/program;f4f26e84-4860-402e-9bcc-a5f1f5e610f1) Sanity content is displayed
3. Verify that CI once again passes


oyvind-stenhaug (approved):
Approved because it's moving in the right direction, though with the following comments:

- I guess title, section contents etc for the program page will be added in a separate PR?
- I'm confused as to why there are so few sessions showing up. I only see one accordion, for San Fransisco, and only two sessions inside (Pause + Reception). But in Sanity there seems to be a lot more of them? (In the "Sessions" tab)
- Presumably "/speaker" in the second bullet point in "Intent" was supposed to be "/program"
- And of course there are frontend issues, but I can fix those in a separate PR if you prefer.

j0nas:
> I guess title, section contents etc for the program page will be added in a separate PR?

Yes. Next step is to delete program.tsx and figure out how to get the program to render through [[...slug]].tsx, being sourced through Sanity.

> I'm confused as to why there are so few sessions showing up. I only see one accordion, for San Fransisco, and only two sessions inside (Pause + Reception). But in Sanity there seems to be a lot more of them? (In the "Sessions" tab)

Based on the way things currently are in Studio, my understanding is that a Program has a `sessions` field which is an array array of "slots", each of which can be associated with a Session, with only the latter being the same as the "Sessions" listed in Studio. So the reason of the lack of content is because the current program doesn't have more sessions associated with it.

> Presumably "/speaker" in the second bullet point in "Intent" was supposed to be "/program"

Yes, thanks, fixed.

> And of course there are frontend issues, but I can fix those in a separate PR if you prefer.

Yes please, per our usual approach. :)

PR #49, opened 2022-02-23
Delete legacy ad-hoc pages

j0nas:
# Delete legacy ad-hoc pages

## Intent

In order to avoid confusion between what's hardcoded and not, this PR deletes all "legacy" ad-hoc pages in _/pages. The intention is that every single page is supposed to be fetched as a "Landing Page" document from Sanity and rendered by _[[...slug]].tsx_.

## Testing this PR

Verify that.. 
1. the logo in the Nav takes you to /, which renders fine
2. That you no longer get a 5XX when accessing /about


oyvind-stenhaug:
@j0nas Oh wait, how can I now work on the Program accordion? This PR deletes that page.

PR #53, opened 2022-02-24
Fix VenuesSection layout

oyvind-stenhaug:
# Fix VenuesSection layout

## Intent

Fix venues section being too wide

## Description

I overlooked this in https://github.com/sanity-io/structured-content-2022/pull/47, it's sort of a regression. On the old /home page, Venues was wrapped in a GridWrapper, but this is missing in the new TextBlock-based regime.

I originally wrote it so that the Venues component itself doesn't have a wrapper, because the Figma sketches looked like the same component might be used in a different context/layout. Perhaps this is not relevant anymore, but in any case the VenuesSection seems like a sensible place to have the wrapper.

## Testing this PR

1. Open the [front page](https://structured-content-2022-web-git-fix-venue-names-section-layout.sanity.build/)
2. Make sure the window is really wide
2. Scroll down and verify that the venue names list ("Oslo HQ • New York […]") isn't wider than the "nav block" of yellow boxes above (the exact way this text renders depends somewhat on the default sans-serif font on your system, so inspecting the `ul` element in the browser's dev tools might also be useful)

### DIFF
--- a/web/components/TextBlock/VenuesSection/VenuesSection.tsx
+++ b/web/components/TextBlock/VenuesSection/VenuesSection.tsx
@@ -1,3 +1,4 @@
+import GridWrapper from '../../GridWrapper';
 import VenueNames from '../../VenueNames';
 
 export const VenuesSection = ({ value: { type, venues } }) => {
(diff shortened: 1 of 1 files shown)


j0nas (approved):
Ah, I didn't notice this, good catch. :+1:

PR #55, opened 2022-02-24
Tickets.tsx: Support `ticketsSection` type section in [[...slug]].tsx

j0nas:
# Tickets.tsx: Support `ticketsSection` type section in [[...slug]].tsx

## Intent

Just like for "Venue" (`venueSection`) and "Sponsor" (`sponsorsSection`), fetch "Tickets" when encountering a `ticketsSection` and pass those along to the associated component. However, unlike the two former entities, tickets are fetched from the "Structured Content 2022" event, which holds the correct ordering. (I will fix querying sponsors and venues from this event too, in a separate PR.) 

## Testing this PR

1. Verify that a pricing table looking sort-of-like the one in the Figma sketches is rendered when viewing [/registration-info](https://structured-content-2022-web-bsoi970wa.sanity.build/registration-info), and that it looks sort-of-right when shrinking viewport to sub-768px width.

### DIFF
--- a/web/components/TextBlock/QuestionAndAnswerCollection/QuestionAndAnswerCollection.tsx
+++ b/web/components/TextBlock/QuestionAndAnswerCollection/QuestionAndAnswerCollection.tsx
@@ -1,11 +1,9 @@
 import Heading from '../../Heading';
 import Block from '../Block';
-import GridWrapper from "../../GridWrapper";
+import GridWrapper from '../../GridWrapper';
 import styles from './QuestionAndAnswerCollection.module.css';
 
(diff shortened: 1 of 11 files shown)


oyvind-stenhaug (approved):
There is a theoretical design issue here in that it's possible to add an arbitrary amount of tickets, while the table display can't really handle that…

Code looks OK to me, though, based on a very superficial skim through. I've pushed a number of commits to adjust the frontend parts, ready for your review.

j0nas:
Your changes look great to me, @oyvind-stenhaug, thank you! I'll merge this now. :)

PR #59, opened 2022-02-25
/registration-info: Tickets section structure update

j0nas:
# /registration-info: Tickets section structure update

## Intent 

Updates the Tickets section component to handle the changes introduced in tonight's Sanity model updates. 

## Description

The logic for figuring out which price is the current was copied from _/studio/schemas/documents/ticket.ts_ (`currentPrice`)

## Testing this PR

- Verify that the table under "Event Pricing" in [/registration-info](https://structured-content-2022-web-jooo8mk6k.sanity.build/registration-info) looks like the Figma sketches, presenting all [Ticket(s) fields available in the Studio](http://localhost:3333/desk/structuredContent2022;58dc88bb-2e29-4de1-8e91-d7755b0afc69%2Ctype%3Dticket%2CparentRefPath%3Dtickets%5B_key%3D%3D%22faac203288c0%22%5D).
- Change the "from" date on the non-"Early bird" "Price and availability" section of the "In-person" Ticket type to a date that's already passed, "Publish" and refresh your browser, and verify that the "Early bird" price for is no longer highlighted, with the 700$ ticket now being the highlighted option. Revert your changes in the Studio.
- [This Shortcut story](https://app.shortcut.com/sanity-io/story/15041/front-end-present-latest-ticket-info-fields-from-studio) says that the ticket prices are not being updated. Verify that this is (no longer?) the case; that updating everything Ticket-related (or at least the price?) works as expected.

@oyvind-stenhaug Commit your CSS and markup changes (if any) to this branch or in a separate PR, whatever is your preference. :slightly_smiling_face:

### DIFF
--- a/web/components/TextBlock/Tickets/Tickets.tsx
+++ b/web/components/TextBlock/Tickets/Tickets.tsx
@@ -7,6 +7,7 @@ import clsx from 'clsx';
 import { Fragment } from 'react';
 import Block from '../Block';
 import { compareAsc, parseISO } from 'date-fns';
+import { PortableText } from '@portabletext/react';
 
 interface TicketsProps {
   value: {
@@ -55,9 +56,7 @@ export const Tickets = ({ value: { type, tickets } }: TicketsProps) => {
                   <div className={styles.name}>{ticket.type}</div>
                   {ticket.description && (
                     <div className={styles.description}>
-                      {ticket.description.map((value) => (
-                        <Block key={value._key} value={value} />
-                      ))}
+                      <PortableText value={ticket.description} />
                     </div>
                   )}
                   <dl className={styles.priceList}>
@@ -137,9 +136,7 @@ export const Tickets = ({ value: { type, tickets } }: TicketsProps) => {
                 <h3 className={styles.name}>{ticket.type}</h3>
                 {ticket.description && (
                   <div className={styles.description}>
-                    {ticket.description.map((value) => (
-                      <Block key={value._key} value={value} />
-                    ))}
+                    <PortableText value={ticket.description} />
                   </div>
                 )}
                 <dl className={styles.priceList}>
--- a/web/components/TextBlock/Tickets/Tickets.tsx
+++ b/web/components/TextBlock/Tickets/Tickets.tsx
@@ -5,7 +5,6 @@ import crossIcon from '../../../images/cross.svg';
 import styles from './Tickets.module.css';
 import clsx from 'clsx';
 import { Fragment } from 'react';
-import Block from '../Block';
 import { compareAsc, parseISO } from 'date-fns';
 import { PortableText } from '@portabletext/react';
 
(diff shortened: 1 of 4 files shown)


web/components/TextBlock/Tickets/Tickets.tsx
```diff
@@ -27,23 +30,59 @@ export const Tickets = ({ value: { type, tickets } }: TicketsProps) => {
...
-            ))}
+            {tickets.map((ticket) => {
+              const currentPrice = currentPriceAndAvailability(ticket);
+              return (
+                <th key={ticket._id} scope="col" className={styles.ticketInfo}>
+                  <div className={styles.name}>{ticket.type}</div>
+                  {ticket.description && (
+                    <div className={styles.description}>
+                      {ticket.description.map((value) => (
+                        <Block key={value._key} value={value} />
```
oyvind-stenhaug, line 59: This is really a part of a more "global" issue (related to the `_span`/`<span>` thing discussed on Slack), so I consider it a nit here.

But it seems to me that we neither support nor need to support any advanced kinds of blocks here, like `RichText` or `Venue`. So it might make sense to render the value more directly.

I wonder if we should use `<PortableText>` directly in cases such as this, or if there should be a simplified version of the `TextBlock.tsx` thing. And I also suspect that the `Block` component means that all inline formatting might be lost, because it doesn't seem to use `<PortableText>` itself.

oyvind-stenhaug (approved):
The list+table duplication is a bit hairy and I see I made some mistakes with it myself earlier, but these adjustments largely look good to me. I've only pushed some smaller tweaks, feel free to comment.

web/components/TextBlock/Tickets/Tickets.tsx
j0nas: Using `<PortableText />` directly here makes sense here, and also looks cleaner. 8fc8688

(I don't even remember what or why `<Block />` is at this point, a small round of cleanup is due after launch. :sweat_smile:)

j0nas:
@oyvind-stenhaug Your changes look very good to me, thank you for the cleanup. If you could give me a thumbs-up on [my last comment](https://github.com/sanity-io/structured-content-2022/pull/59#discussion_r814814594) here, I'll merge this. :slightly_smiling_face:

web/components/TextBlock/Tickets/Tickets.tsx
```diff
@@ -4,6 +4,10 @@ import checkmarkIcon from '../../../images/checkmark.svg';
 import crossIcon from '../../../images/cross.svg';
 import styles from './Tickets.module.css';
 import clsx from 'clsx';
+import { Fragment } from 'react';
+import Block from '../Block';
```
oyvind-stenhaug, line 8: I don't know why the linter doesn't flag this as an unused import, but it seems to be unused now…

```suggestion
```

oyvind-stenhaug (approved):
Thanks, looks good. I think the Block import is now superfluous, but other than that I have no further comments :)

PR #62, opened 2022-02-25
Header menu selected state

oyvind-stenhaug:
# Header menu selected state

## Intent

Underline regular menu items in the tablet/desktop header menu when they are "selected" i.e. match the current page ([Shortcut story 15079](https://app.shortcut.com/sanity-io/story/15079/selected-states-for-main-navs-underline))

## Description

Underline is also used as a hover style here. Got a clarification from Celine in Slack that it's OK to keep both behaviors (so I would consider it expected that a menu item is underlined if it is hovered _and/or_ its href matches the currently opened page). Moreover that this "selected" state is not used for the mobile menu (which doesn't look like a menu bar).

With the introduction of the `currentPath` prop on `Nav`, the `onFrontPage` prop looks somewhat redundant. I chose to keep it because—as currently written—the `MyApp` component (`_app.tsx`) needs to know that boolean itself anyway.

## Testing this PR
1. Open the [About page](https://structured-content-2022-web-git-header-menu-selected-state.sanity.build/about)
2. Make sure the window is wide enough that the header menu bar is displayed at the top of the screen, and verify that "About" in that menu is underlined
3. Click "Sponsorship" in the menu and verify that this word is now underlined (even when not hovered)

### DIFF
--- a/web/components/Nav/Nav.tsx
+++ b/web/components/Nav/Nav.tsx
@@ -21,6 +21,22 @@ export const Nav = ({ onFrontPage }: NavProps) => {
   const toggleMenu = () => setMenuOpened(!menuOpened);
   const closeMenu = () => setMenuOpened(false);
 
+  const BasicMenuItem = ({
+    urlPath,
+    label,
+  }: {
+    urlPath: string;
+    label: string;
+  }) => (
+    <li>
+      <Link href={urlPath}>
+        <a className={styles.link} onClick={closeMenu}>
+          {label}
+        </a>
+      </Link>
+    </li>
+  );
+
   return (
     <nav className={clsx(styles.nav, onFrontPage && styles.onFrontPage)}>
       <GridWrapper>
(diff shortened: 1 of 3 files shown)


web/components/Nav/Nav.tsx
```diff
@@ -21,6 +22,28 @@ export const Nav = ({ onFrontPage }: NavProps) => {
...
   const closeMenu = () => setMenuOpened(false);
 
+  const BasicMenuItem = ({
+    urlPath,
+    label,
+  }: {
+    urlPath: string;
+    label: string;
+  }) => {
+    const isCurrentPage = urlPath === currentPath;
```
j0nas, line 32: Nit(personal preference): this constant might be inlined, which allows dropping the surrounding curlies and the explicit `return`.

j0nas, line 32: Nit(personal preference): this constant could be inlined, which allows dropping the surrounding curlies and the explicit `return`.

web/components/Nav/Nav.tsx
```diff
@@ -21,6 +22,28 @@ export const Nav = ({ onFrontPage }: NavProps) => {
   const toggleMenu = () => setMenuOpened(!menuOpened);
   const closeMenu = () => setMenuOpened(false);
 
+  const BasicMenuItem = ({
```
j0nas, line 25: Nice refactor! :100:

j0nas (approved):
LGTM!

web/components/Nav/Nav.tsx
oyvind-stenhaug: Yep, I actually had it like that initially. I have a slight preference to factoring out const/variable declarations to give them useful names, but it's not much of a difference here so I don't mind inlining it. (Number of lines is the same, there's more wrapping with the inlined constant.)

oyvind-stenhaug: 1f1d8d248bb122d2a76e1daa783be7fe83721c31

PR #65, opened 2022-02-28
Hero summary, CTA

oyvind-stenhaug:
# Hero summary, CTA

## Intent

Add support for showing a summary paragraph and/or a "call to action" in a page's "hero" (page-specific header).

## Description

Here I've tried to reuse the TypeScript definition for the Sanity fields in the component itself. Upside is that it's less duplication, but downsides might be that you have to go looking in some other file for the type, plus increased coupling between components and Sanity fields.

Also not sure about how to handle the `PortableTextComponentProps` issue (see [Vercel build of a992e90](https://vercel.com/sanity-io/structured-content-2022-web/7TQzhTmZS7wEFPTwtKMTSVewupNa) which failed). Right now I've split it into two: one component that exists to be used as a "Portable Text component", and one "plain" component which can be used elsewhere (and the former just forwards to the latter). Could be a bit confusing…

I left the `name` field since I can still see it in the [staging studio](https://stagingadmin.structuredcontent.live/) (which doesn't seem to be updated to the latest staging commit?) and some pages still seem to have it, such as /venues. Could remove support for it completely, though.

## Testing this PR
1. Open [/registration-info](https://structured-content-2022-web-63p68j6ki.sanity.build/registration-info) (which at the time of writing has contents for the summary and CTA)
2. Verify that the "hero" section at the top of the page looks OK, with header, intro and "button" leading to www.hopin.com

### DIFF
--- a/web/components/Hero/Hero.module.css
+++ b/web/components/Hero/Hero.module.css
@@ -4,6 +4,13 @@
   padding: 40px 0 80px;
 }
 
+.summary {
+  font: var(--font-body-small-medium);
+  letter-spacing: var(--letter-spacing-body-small);
+  margin: 32px auto 0 auto;
+  max-width: var(--max-body-text-width);
+}
+
 .cta {
   margin-top: 32px;
 }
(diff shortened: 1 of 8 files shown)


oyvind-stenhaug:
Shortcut: https://app.shortcut.com/sanity-io/story/14919/implement-hero-design-title-ingress-cta

j0nas (approved):
Changes look good to me, good job.

PR #67, opened 2022-02-28
Editorial articles

j0nas:
# Editorial articles

## Intent

Render (to-be-styled) [Editorial Articles](http://localhost:3333/desk/article;febd810c-706d-46e2-a391-3e910fc4b06c), e.g. Code of Conduct on  the `/articles/:slug` path.

## RFC

I opted for a separate route rather than stuffing more content into _[[...slug]].tsx_, but the latter might also be an option. Thoughts?

## Testing this PR

Go to [/article/code-of-conduct](https://structured-content-2022-web-35sdjq4pa.sanity.build/article/code-of-conduct) and verify that the content of the respective [Editorial Article](http://localhost:3333/desk/article;febd810c-706d-46e2-a391-3e910fc4b06c) renders.

### DIFF
--- a/web/components/Footer/Footer.module.css
+++ b/web/components/Footer/Footer.module.css
@@ -22,6 +22,7 @@
 
 .social {
   margin: 38px 0;
+  padding: 0;
   text-align: center;
 }
 
@@ -49,6 +50,7 @@
 .links {
   list-style: none;
   margin: 20px -12px 0;
+  padding: 0;
   text-align: center;
 }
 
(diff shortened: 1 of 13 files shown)


oyvind-stenhaug (approved):
@j0nas I've adjusted the styling here, mostly by tweaking the "global" base styling so we don't need to do any custom "serialization" of the Portable Text. Also copy/pasting layout styles from RichText as discussed on Slack.

I think your choice of creating `article/[slug].tsx` makes sense at this stage, limiting the complexity of `[[...slug.tsx]]` somewhat.

PR #69, opened 2022-03-01
Fix internal Hero CTA

oyvind-stenhaug:
# Fix internal Hero CTA

## Intent

Fix rendering of "Call to action" in a landing page's "Hero" component, in the case where that CTA is referencing a "Route" document rather than specifying a URL

## Description

Was missing some dereferencing in the query here.

Also related to this CTA is the fact that it could be pointing to an `article` instead of a `page`. It seems it's currently unclear how slugs on these articles should work (see Slack threads e.g. on our [shared channel with Behalf](https://wja.slack.com/archives/C02E2ERKR7E/p1646129047370819)). Depending on how that turns out, the SimpleCallToAction component may need some sort of helper function to handle converting the slug into a URL, along the lines of web/util/entityPaths.ts.

Because of this uncertainty, the latter problem is not handled in this PR.

## Testing this PR
1. Open the [Venues](https://structured-content-2022-web-qp8p6f38f.sanity.build/venues) page
2. Verify that there is a "Get registration info" button (link) right below the page summary (topmost yellow section)

### DIFF
--- a/web/pages/[[...slug]].tsx
+++ b/web/pages/[[...slug]].tsx
@@ -21,7 +21,13 @@ const QUERY = groq`
       ...,
       page-> {
         name,
-        hero,
+        hero {
+          ...,
+          callToAction {
+            ...,
+            reference->{slug},
+          },
+        },
         sections[] {
           _type == 'reference' => @-> {
             sections[] {


web/pages/[[...slug]].tsx
```diff
@@ -21,7 +21,13 @@ const QUERY = groq`
       ...,
       page-> {
         name,
-        hero,
+        hero {
+          ...,
+          callToAction {
+            ...,
+            reference->{slug},
```
Comment (j0nas) [inline web/pages/[[...slug]].tsx:28]:
```suggestion
            reference->,
```

Nit: I've tried to refrain from nesting the queries more than necessary. This suggestion includes more than just the `slug` field in the response now, if present, whereas the previous alternative would only include `slug`.

j0nas (approved):

web/pages/[[...slug]].tsx
Comment (oyvind-stenhaug) [reply web/pages/[[...slug]].tsx:28]: Oh OK, I thought this was just done due to lack of time.

Only including `slug` is exactly what I was going for, because that's all we need (as per the `SimpleCallToAction` TS type). As far as I know, all extra/unused fields we query for here will add to the size of the `pageProps` JSON that is loaded client-side when navigating between pages.

https://www.sanity.io/blog/3-simple-things-in-groq-to-supercharge-your-frontend-development#db26fcd247c4

If you nevertheless prefer to leave the projection out (e.g. for consistency), I'll commit your suggestion, though. Just checking :)

Comment (oyvind-stenhaug) [reply web/pages/[[...slug]].tsx:28]: Yeah it's a bit less convenient for the phase where everything is very unstable, as that blog article says… :)

PR #70, opened 2022-03-01
Meta tags for SEO

j0nas:
# Meta tags for SEO

## Intent

Add basic meta tags to `<head>` using [next-seo](https://github.com/garmeeh/next-seo).

## Testing this PR

1. Go to the [frontpage](https://structured-content-2022-web-1xujtju6z.sanity.build/)
2. Verify that we render all the meta tags we can, given the `seo` fields available in Studio for the [front page Route](https://stagingadmin.structuredcontent.live/desk/route;9e8a9f97-2573-493d-aae9-3f5d8289ad58)
    * (Are there other essential tags we should render that we currently don't?)
3. Go to [/shared-sections-test](https://structured-content-2022-web-1xujtju6z.sanity.build/shared-sections-test)
4. Verify that a `noindex` meta tag is present in the document's `<head>` corresponding to the [Studio "Hide this route" setting](https://stagingadmin.structuredcontent.live/desk/route;36c8bda6-e3a3-4671-a260-cdaf3dcd2e43) being enabled for the Route

### DIFF
--- a/web/pages/[[...slug]].tsx
+++ b/web/pages/[[...slug]].tsx
@@ -1,7 +1,7 @@
 import clsx from 'clsx';
 import { groq } from 'next-sanity';
-import { NextSeo } from "next-seo";
-import urlJoin from "proper-url-join";
+import { NextSeo } from 'next-seo';
+import urlJoin from 'proper-url-join';
 import { useEffect, useState } from 'react';
 import Hero from '../components/Hero';
 import TextBlock from '../components/TextBlock';
@@ -93,14 +93,14 @@ interface RouteProps {
         sections: Section[];
       };
       seo: {
-        _type: 'seo',
+        _type: 'seo';
         title: string;
         description: string;
         image?: {
           asset: {
             url: string;
-          }
-        },
+          };
+        };
         noIndex?: boolean;
       };
     };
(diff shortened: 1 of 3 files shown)


web/pages/[[...slug]].tsx
```diff
@@ -150,6 +164,13 @@ const Route = ({
 
   return (
     <>
+      <NextSeo
+        title={title}
+        description={seoDescription}
+        canonical={urlJoin('https://structuredcontent.live', currentPath)}
+        noindex={noIndex}
+        openGraph={image ? { images: [{ url: image?.asset?.url }] } : undefined}
```
Comment (oyvind-stenhaug) [inline web/pages/[[...slug]].tsx:172]: I tried to test this with your shared-sections-test route, it doesn't seem to work: `<meta property="og:image"/>` with "View Source" on https://structured-content-2022-web-1xujtju6z.sanity.build/shared-sections-test.

Nit: I'm not sure, but I am guessing that the `.url` will just give a URL to the original image, however huge it may be. Ideally, I would think that it ought to be [transformed](https://www.sanity.io/docs/image-urls) in some way to fit or clip it within some limit. E.g. [Facebook requires](https://developers.facebook.com/docs/sharing/webmasters/images/) the image to be max 8 MB. 1260x630 might be a good max size…

oyvind-stenhaug (changes requested):
I'm not familiar with NextSeo and would probably find it simpler to specify the elements directly (WYSIWYG), but it's your call.

It doesn't look like the image works, though?

web/pages/[[...slug]].tsx
Comment (j0nas) [reply web/pages/[[...slug]].tsx]: Good catch, I didn't think to add a test image to the test route.. 29411fc should address these concerns.

oyvind-stenhaug (approved):

PR #72, opened 2022-03-01
"Nav block" shapes

oyvind-stenhaug:
# "Nav block" shapes

## Intent

Implement the updated design of the "Nav block", with changed number/distribution of empty boxes and with various custom shapes to make it look less like a brick wall

## Description

Shortcut: https://app.shortcut.com/sanity-io/story/15395/implement-new-design-for-nav-block

I felt that writing these shapes as SVG was the simplest way to solve it, rather than adding even more empty HTML elements and styling those in different ways.

As Sindre mentioned in the Shortcut story, SVGs exported the "normal" way will end up looking stretched in ways we don't want, with the border radiuses being stretched non-uniformly. This is basically because such SVGs will have a "natural size"/aspect ratio.

Instead, I added some hand-written SVGs that really turned out to be quite quick to make. They don't have any intrinsic/natural size, so the rounded corners aren't scaled up or down. But because I use percentages for positioning/widths etc, the SVGs still fill the area allocated to them (which is basically 100% of the background area of the given `FakeItem` `<li>`).

As discussed before we could consider inlining the SVGs as `data:` URIs if that is an improvement for performance, but I considered this out of scope for the PR since we don't have a general setup for doing this, as far as I know. (Unless we manually inline them in the CSS source. This is definitely doable just not very pretty/maintainable.)

(We also have an issue for this, in GitHub at least. I could add it to Shortcut too.)

## Testing this PR
1. Open https://structured-content-2022-web-git-nav-block-shapes.sanity.build/
2. Verify that the "nav block" of black-on-orange links (Program, Speakers etc) has various shapes in it other than plain rounded-corner rectangles, and corresponds to the "Approved" Figma sketches labelled "Homepage v2"

### DIFF
--- a/web/components/NavBlock/NavBlock.module.css
+++ b/web/components/NavBlock/NavBlock.module.css
@@ -50,6 +50,10 @@
   background: url('../../images/navblock_plus.svg') 100% 100% no-repeat;
 }
 
+.fakeItem.shapeC {
+  background: url('../../images/navblock_c.svg') 100% 100% no-repeat;
+}
+
 .fakeItem.mobile {
   display: block;
 }
--- a/web/components/NavBlock/NavBlock.module.css
+++ b/web/components/NavBlock/NavBlock.module.css
@@ -107,6 +107,10 @@
     padding: 28.5px 39px;
   }
 
+  .fakeItem {
+    border-radius: 12px;
+  }
+
   .fakeItem.tablet,
   .divider.tablet {
     display: none;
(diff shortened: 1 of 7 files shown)


j0nas (approved):
I'm amazed at how the end result you arrived at looks _exactly the same_ as the Figma sketches. Good job!

oyvind-stenhaug:
Thank you!

PR #78, opened 2022-03-02
Editorial articles accessible from root

j0nas:
# Editorial articles accessible from root 

## Intent

Solves [this](https://app.shortcut.com/sanity-io/story/15481/routing-update-to-handle-editorial-articles) Shortcut Story, which calls for Editorial Articles being accessible from both _/:slug_ and _/article/:slug_ routes.

## Description

On startup, we now fetch all available `article` documents and add them to Next.js' [`rewrite`](https://nextjs.org/docs/api-reference/next.config.js/rewrites) config. 

I don't think this array is ever updated post-deployment, so [the relevant Shortcut Story](https://app.shortcut.com/sanity-io/story/14538/rebuild-page-when-respective-sanity-content-is-updated) should also update this setting if any article is added/removed. 

## Testing this PR

1. Verify that the build logs for this branch log (three, at the time of writing) rewrites being added 
2. Visit an article from root, e.g. [/code-of-conduct](https://structured-content-2022-web-m5ams3t6l.sanity.build/code-of-conduct) and verify that you're presented with the same content as if you'd accessed [/article/code-of-conduct](https://structured-content-2022-web-m5ams3t6l.sanity.build/article/code-of-conduct)
3. Verify that the `canonical` meta tag points to `https://structuredcontent.live/article/code-of-conduct`, including the `/article` bit
4. Verify that meta tags render correctly when accessing articles both from _/:slug_ and _/article/:slug_

### DIFF
--- a/web/components/MetaTags/MetaTags.tsx
+++ b/web/components/MetaTags/MetaTags.tsx
@@ -1,0 +1,52 @@
+import { NextSeo } from 'next-seo';
+import urlJoin from 'proper-url-join';
+import { imageUrlFor } from '../../lib/sanity';
+
+interface MetaTagsProps {
+  title: string;
+  description: string;
+  image?: object;
+  currentPath: string;
+  noIndex?: boolean;
+  rewrittenArticleSlugs?: string[];
+}
+
+export const MetaTags = ({
+  title,
+  description,
+  image,
+  currentPath,
+  noIndex,
+  rewrittenArticleSlugs,
+}: MetaTagsProps) => {
+  const isRewrittenPath =
+    Array.isArray(rewrittenArticleSlugs) &&
+    rewrittenArticleSlugs.includes(
+      urlJoin(currentPath, { leadingSlash: false })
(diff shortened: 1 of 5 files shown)


oyvind-stenhaug (changes requested):
As commented in Shortcut, I think this rewrite means we need to account for it in the canonical URL meta information? That's the whole point of `canonical` after all

j0nas:
> As commented in Shortcut, I think this rewrite means we need to account for it in the canonical URL meta information? That's the whole point of `canonical` after all

@oyvind-stenhaug I've fixed this in 9dd89df & 010cb48 and updated the testing steps with an additional step. Thanks for reminding me to fix this. :slightly_smiling_face:

oyvind-stenhaug (approved):
It works :+1: 

Looks like the meta description comes from the article's own Summary and not the SEO section on the route, but I don't know if that's expected or out of scope for this PR. It's kind of a confusing setup…

PR #81, opened 2022-03-03
Allow paragraphs within `textAndImage` section

j0nas:
# Allow paragraphs within `textAndImage` section

## Intent

- Fixes [this](https://app.shortcut.com/sanity-io/story/15521/allow-paragraphs-within-the-value-proposition-module) Story
- Fixes typings for the `<TextAndImage />`  section component

## Testing this PR

Verify that the "Value Proposition" section on the [frontpage](https://structured-content-2022-web-486p1lyi6.sanity.build/) is now divided into paragraphs.


oyvind-stenhaug (changes requested):
Looks OK given the circumstances of how TextBlock is currently set up.

Just one minor thing: the added quotation marks are inconsistent with the existing ones. Maybe run the two files through Prettier? (Reason this isn't a nit is that such reformatting tends to happen at some point anyway, and make the diff in some later PR end up bigger/more cumbersome to sift through)

Also, note that `<Block>` is also used in `QuestionAndAnswerCollection` which thus suffers the same problem of missing formatting. See e.g. "What do in-person tickets get me?" on /registration-info currently. Up to you if this is treated as a separate issue/PR, or is done here so that Block can be deleted.

j0nas:
@oyvind-stenhaug I was initially skeptical towards deleting `<Block />` because I wasn't sure if it was still used by the Sanity model, but quickly browsing through the pages after deleting it yields no warnings, so it looks like your suggestion was the right call. efa278a, f215263.

oyvind-stenhaug:
My understanding is that the presence of the `block` type in an array is what makes it be "Portable Text". E.g. an array of `string` is not portable text, and one wouldn't use `<PortableText>` to render it. But a `type: "array", of: [{ type: "block"}]` _is_ Portable Text, and that's what `<PortableText>`/`@portabletext/react` is for; the _library_ handles the display of those `block` items, which potentially has `span` children, `marks` to specify formatting such as bold and italic, and all that jazz.

What complicates the issue is that you can also have items of other types _alongside_ the `block` ones, such as for `studio/schemas/arrays/simpleBlockContent.ts`. Or you can specify an "[array of inline content types](https://www.sanity.io/docs/block-type#of-d0f97ffa1dd9)" that are custom types you can nest inside your rich text. _Those_ types are not handled by the Portable Text library, so we need to tell it which components to use.

As discussed in Slack I wouldn't have used `<PortableText>` to render all sorts of Sanity schema types, only the ones that are "rich text" aka "block content" aka "Portable Text". I.e. basically `simpleBlockContent` and `blockContent` in this case. But here we are, it's too late to change now even if I could convince you (at least before launch) :) I am confident that removing custom rendering of `block` is the right thing, though!

oyvind-stenhaug (approved):
Looks good to me :+1:

PR #82, opened 2022-03-03
Fallback OpenGraph image meta tags

j0nas:
# Fallback OpenGraph image meta tags

## Intent

- Fixes [this](https://app.shortcut.com/sanity-io/story/15577/handle-open-graph-image) Story.
- Creates `Figure` type and replaces all inline Figure declarations with references to the new type
- Outputs `og:image:width` and `og:image:height` meta tags for the image

## Description

In my browser, the image linked to in the meta tags for [/shared-sections-test][deploy-test] is 630x630, and we don't have the dimension data available, so we can't output `og:image:width`/-`height` tags for non-fallback images. There is [an issue](https://github.com/sanity-io/sanity/issues/1627) for this for the `createImageUrlBuilder`. ~Maybe omitting the tags is better?~ I did so in 72ad10a49ff2be8a0f569e37ed7585a6aa7f6916.

## Testing this PR

1. Visit the [frontpage][deploy]
2. Verify that there are three `og:image` meta tags, and that they have the correct values corresponding to the fallback image's attributes.
3. Visit [/shared-sections-test][deploy-test]
4. Verify that the meta tags hold the info of [the Route's](https://stagingadmin.structuredcontent.live/desk/route;36c8bda6-e3a3-4671-a260-cdaf3dcd2e43) associated image, not the fallback image's

[deploy]: https://structured-content-2022-web-j60fk11gg.sanity.build/
[deploy-test]: https://structured-content-2022-web-j60fk11gg.sanity.build/


web/components/MetaTags/MetaTags.tsx
```diff
@@ -33,22 +35,25 @@ export const MetaTags = ({
...
+          image
+            ? {
+                url: imageUrlFor(image)
+                  .ignoreImageParams()
+                  .size(1260, 630)
+                  .url(),
+                alt: image.alt,
+              }
+            : {
+                url: '/static/images/opengraph-image.svg',
```
oyvind-stenhaug, line 49: Referring to item 2 in the previous comment: I believe this should be `opengraphImage.url` instead, in order to use the automatically generated "cache-breaking" URL (once the file is moved out of `static`).

web/types/Figure.ts
```diff
@@ -0,0 +1,10 @@
+export type Figure = {
```
oyvind-stenhaug, line 1: Nice refactor :+1:

oyvind-stenhaug (changes requested):

web/components/MetaTags/MetaTags.tsx
```diff
@@ -1,11 +1,13 @@
 import { NextSeo } from 'next-seo';
 import urlJoin from 'proper-url-join';
+import opengraphImage from '../../public/static/images/opengraph-image.svg';
```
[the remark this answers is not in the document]
j0nas: Good catch! I find myself taking SVG support for granted too often :face_exhaling: 9ee794a.

web/components/MetaTags/MetaTags.tsx
j0nas: Of course! I couldn't immediately find out how to get such an URL, but of course next supports in in the form of `.src`. :100: Also fixed in 9ee794a.

oyvind-stenhaug (approved):
Nit: Not sure if the image was intended to be 1260px _tall_ instead of 1260px _wide_ like the Sanity image would be (if present)?

In terms of file size it's quite small, though (~55 kB), as one might expect given the nature of the logo (big areas with uniform flat color). So just a nit.

oyvind-stenhaug:
Oh, by the way, regarding your comment on `og:image:width` and `og:image:height`: This sounds closely related to the issue we have on [using Sanity image dimensions to limit "cumulative layout shift"](https://app.shortcut.com/sanity-io/story/15073/use-sanity-image-dimensions-in-html-performance). In that "Story", I referred to `@sanity-io/asset-utils` as a potentially relevant library. Only based on a quick web search, though.

oyvind-stenhaug:
Added a comment to that story about the `og:` case.

j0nas:
@oyvind-stenhaug Updated image to 1260x630. :slightly_smiling_face:

PR #84, opened 2022-03-03
Program venue updates

oyvind-stenhaug:
# Program venue updates

## Intent

Update Program section to use the program venue's timezone when listing out the sessions, reflecting this in the program's heading. See [Shortcut story "Update program page"](https://app.shortcut.com/sanity-io/story/15643/update-program-page).

## Description

This is using date-fns-tz, just like an earlier [PR for Tickets](https://github.com/sanity-io/structured-content-2022/pull/80). Timezone is defined on a venue in Sanity. As discussed in video meeting with Sanity yesterday, we can assume that a program will only have one venue (even though their data model supports multiple).

Sketches in Figma display the time zone as a long-form location-independent string, so e.g. "British Summer Time", not "Europe/London". This representation varies based on time. In the edge case where the program spans two different such time zone representations (due to daylight saving time switch), the program heading just shows the timezone that applies at the program's start. I figure that's close enough.

As requested in the Shortcut story, this PR also changes the heading from using the internal program name to using the (first) venue's name (which is shorter). It does not include solving the order of programs, however, since I think this requires some sort of sorting data to be specified in the schema, which would be up to Knut. I'll ping Celine about that.

## Testing this PR
1. Open the [Program page](https://structured-content-2022-web-git-program-venue-updates.sanity.build/program)
2. Verify that the headings include venue name and timezone, e.g. "New York - Eastern Daylight Time" and not "New York City" (or "Oslo 2022")
3. Verify that the days and times specified in each program are given in the indicated timezone, e.g. PDT for the San Fransisco sessions. (I used "Inspect" in the studio for this purpose to see the exact timestamp that was given for `startDateTime`, and [timeanddate.com](https://www.timeanddate.com/time/zones/pdt) to see which offsets the American cities are at. Or, as far as I can see, if you look at the displayed date+time in Sanity, that seems to be using your local timezone.)

### DIFF
--- a/web/components/TextBlock/Programs/Programs.tsx
+++ b/web/components/TextBlock/Programs/Programs.tsx
@@ -45,7 +45,7 @@ export const Programs = ({
               content: (
                 <div key={program._id}>
                   <h3 className={clsx(styles.dayHeader, styles.first)}>
-                    {formatDateWithDay(program.startDateTime)}
+                    {formatDateWithDay(program.startDateTime, timezone)}
                   </h3>
                   {program.sessions.map((session, index) => {
                     const Session = (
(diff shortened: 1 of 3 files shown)


j0nas (approved):
I can't test this since the Program section is gone, but the code looks good to me. Good job!

PR #85, opened 2022-03-04
Programs section: support 'highlighted' type

j0nas:
# Programs section: support 'highlighted' type

## Intent

- Renders `highlighted` Programs.
- Solves [this](https://app.shortcut.com/sanity-io/story/15741/programs-section-not-updated-to-front-end-at-the-moment) Shortcut Story.

## Testing this PR

1. Visit [/program](https://structured-content-2022-web-c38lo1b1g.sanity.build/program)
2. Verify that the Programs section renders, and that there are no console warnings about anything unrecognized

### DIFF
--- a/web/components/TextBlock/Programs/Programs.tsx
+++ b/web/components/TextBlock/Programs/Programs.tsx
@@ -12,15 +12,16 @@ import {
 import Accordion from '../../Accordion';
 import GridWrapper from '../../GridWrapper';
 import styles from './Programs.module.css';
+import { getCollectionForSelectionType } from "../../../util/entity";
 
 type ProgramsProps = {
   type: EntitySectionSelection;
-  programs?: Program[];
   allPrograms: Program[];
+  programs?: Program[];
 };
 
 export const Programs = ({
-  value: { type, programs, allPrograms },
+  value: { type, allPrograms, programs },
   index,
 }: PortableTextComponentProps<ProgramsProps>) => {
   if (type === 'all' || type === 'highlighted') {
@@ -28,7 +29,7 @@ export const Programs = ({
       <GridWrapper>
         <Accordion
           baseId={`accordion-${index}`}
-          items={(programs || allPrograms).map((program) => {
+          items={getCollectionForSelectionType(type, allPrograms, programs).map((program) => {
             const firstVenue = program?.venues[0];
             const programName = firstVenue?.name || program.internalName;
             const timezone = firstVenue?.timezone || 'UTC';
--- a/web/components/TextBlock/Programs/Programs.tsx
+++ b/web/components/TextBlock/Programs/Programs.tsx
@@ -12,7 +12,7 @@ import {
 import Accordion from '../../Accordion';
 import GridWrapper from '../../GridWrapper';
 import styles from './Programs.module.css';
-import { getCollectionForSelectionType } from "../../../util/entity";
+import { getCollectionForSelectionType } from '../../../util/entity';
 
 type ProgramsProps = {
   type: EntitySectionSelection;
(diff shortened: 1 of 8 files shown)


web/components/TextBlock/Programs/Programs.tsx
```diff
@@ -15,19 +15,20 @@ import styles from './Programs.module.css';
...
   allPrograms: Program[];
 };
 
 export const Programs = ({
-  value: { type, allPrograms },
+  value: { type, programs, allPrograms },
   index,
 }: PortableTextComponentProps<ProgramsProps>) => {
-  if (type === 'all') {
+  if (type === 'all' || type === 'highlighted') {
```
oyvind-stenhaug, line 26: Nit: This doesn't work properly in the following scenario:

1. (In Sanity Studio) Create a programs section with type 'highlighted' and select the Oslo program. Publish.
2. (In the app) Programs section now shows Oslo only, as expected.
3. (In Sanity Studio) Change the programs section type from 'highlighted' to 'all'. Publish.
4. (In the app) Programs section now still only shows Oslo

This is currently demonstrated on the /shared-sections-test page.

The studio doesn't clear out the `programs` array, it just hides it. Which might theoretically be a feature if you want to switch back from "all" to the old "highlighted" configuration later on, I suppose.

Suggestion part 1/2:
```suggestion
  if (type === 'all' || type === 'highlighted') {
    const selectedPrograms = type === 'all' ? allPrograms : programs;
```

web/components/TextBlock/Programs/Programs.tsx
```diff
@@ -15,19 +15,20 @@ import styles from './Programs.module.css';
...
     return (
       <GridWrapper>
         <Accordion
           baseId={`accordion-${index}`}
-          items={allPrograms.map((program) => {
+          items={(programs || allPrograms).map((program) => {
```
oyvind-stenhaug, line 31: Suggestion part 2/2:
```suggestion
          items={selectedPrograms.map((program) => {
```

oyvind-stenhaug (approved):
Thanks for the speedy fix, and sorry about the slow review :)

Left a couple of comments about the all/highlighted logic. Both marked as nit because it's better to have 'highlighted' working in the simple case than for it to not be working at all, now that Sanity started using it on the Programs page.

web/components/TextBlock/Programs/Programs.tsx
j0nas: Interesting catch! What do you think about eb14b33 and its application in b5e4dbf? This was applied to the sections where it was straightforward to do so; a larger-scale unification is probably in order to streamline using this in some sort of `<EntitySection />` component.

web/pages/[[...slug]].tsx
```diff
@@ -96,6 +96,14 @@ const QUERY = groq`
             },
             _type == "programsSection" => {
               ...,
+              programs[]-> {
```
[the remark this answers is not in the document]
Comment (j0nas) [reply web/pages/[[...slug]].tsx:99]: Good suggestion. I filed #91 for this in order not to bloat this PR further given the more-extensive-than-expected changes introduced as part of addressing the comment below.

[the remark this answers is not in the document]
Comment (oyvind-stenhaug) [reply web/pages/[[...slug]].tsx:99]: :+1:

web/components/TextBlock/VenuesSection/VenuesSection.tsx
```diff
@@ -20,26 +21,9 @@ export const VenuesSection = ({
...
-
-    default:
-      console.error(`Unrecognized VenuesSection type: '${type}'`);
-      return null;
-  }
+  return (
+    <GridWrapper>
+      <VenueNames venues={getCollectionForSelectionType(type, allVenues, venues)} />
+    </GridWrapper>
+  );
```
oyvind-stenhaug, line 30: This is simplified a tad too much. I would prefer that no empty `GridWrapper` elements are output to the DOM when there are no venues. So either keep the `none` special case from before, just like what Programs is doing right now even after the refactor, or (maybe even better) make this a conditional return based on the return value of `getCollectionForSelectionType` being non-empty.

E.g. something along the lines of (not tested)

```suggestion
  const selectedVenues = getCollectionForSelectionType(type, allVenues, venues);
  return selectedVenues.length > 0 && (
    <GridWrapper>
      <VenueNames venues={selectedVenues} />
    </GridWrapper>
  );
```

oyvind-stenhaug (changes requested):

web/components/TextBlock/Programs/Programs.tsx
oyvind-stenhaug: I think that looks OK, at least for now. I would think that if https://github.com/sanity-io/structured-content-2022/issues/91 is solved then it would no longer be relevant, as we would avoid fetching the irrelevant collection from Sanity in the first place.

web/components/TextBlock/VenuesSection/VenuesSection.tsx
j0nas: Fixed in b129cc9.

oyvind-stenhaug: Thanks!

oyvind-stenhaug (approved):

PR #87, opened 2022-03-04
Hardcoded conference dates

oyvind-stenhaug:
# Hardcoded conference dates

## Intent

Replace the current UTC display of the conference start/end timestamps (which were fetched from Sanity) with two hardcoded sets of start/end dates for the Structured Content 2022 conference specifically—one line for US and one for UK&Europe. As specified in [Shortcut](https://app.shortcut.com/sanity-io/story/14897/home-page-presentation-of-dates-across-timezones)

## Description

As discussed in the Shortcut story (and way back on video with Knut), presenting such semi-localized dates based on a single pair of start/end timestamps isn't necessarily realistic (does the conference really start and end simultaneously world-wide, or is it actually a moving "window"…) Furthermore, according to the story, this conference might be an exception, so it has been decided that we'll hardcode it for now and possibly revisit later.

Because of the increased lengths of these strings, the font size has been turned down as per the Figma sketches.

## Testing this PR
1. Open the [front page](https://structured-content-2022-web-bc2dmjic9.sanity.build/)
2. Verify that the start/end dates shown directly below the big conference logo are now shown in two separate paragraphs, one for "US" and one for "UK & Europe", as per the sketches in Shortcut/Figma

### DIFF
--- a/web/components/ConferenceHeader/ConferenceHeader.tsx
+++ b/web/components/ConferenceHeader/ConferenceHeader.tsx
@@ -29,9 +29,8 @@ export const ConferenceHeader = ({
       </h1>
       <div className={styles.summary}>
         <div className={styles.datesAndHostedByWrapper}>
-          <p className={styles.dates}>
-            <DateRange startTimestamp={startDate} endTimestamp={endDate} />
-          </p>
+          <p className={styles.dates}>US: May 24–25, 2022</p>
+          <p className={styles.dates}>UK &amp; Europe: May 25–26, 2022</p>
           <p className={styles.hostedBy}>Hosted by Sanity</p>
         </div>
         <p className={styles.description}>{description}</p>
--- a/web/components/ConferenceHeader/ConferenceHeader.tsx
+++ b/web/components/ConferenceHeader/ConferenceHeader.tsx
@@ -3,15 +3,11 @@ import styles from './ConferenceHeader.module.css';
 
 interface ConferenceHeaderProps {
   name?: string;
-  startDate?: string;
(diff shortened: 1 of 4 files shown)


web/components/ConferenceHeader/ConferenceHeader.tsx
```diff
@@ -29,9 +29,8 @@ export const ConferenceHeader = ({
       </h1>
       <div className={styles.summary}>
         <div className={styles.datesAndHostedByWrapper}>
-          <p className={styles.dates}>
-            <DateRange startTimestamp={startDate} endTimestamp={endDate} />
-          </p>
```
j0nas, line 34: Unused props and imports here now; these can also be removed from the Sanity query now and [[...slug]].tsx can stop passing these to the component.

j0nas (approved):
Looks good. There's some code that can be deleted now; approved to merge this PR once that's taken care of. :+1:

web/components/ConferenceHeader/ConferenceHeader.tsx
oyvind-stenhaug: Good point! I figured I'd leave the DateRange implementation as-is because some version of it might be needed in the future once the temporary hardcoding can be fixed, but at the very least it makes complete sense not to spend bandwidth pointlessly fetching the unused data. And I should've removed the import as well.

- Import removed in 691eba3a1a3f1a1fbf3c7a418b2853c596d233ee
- startDate/endDate removed from query/props in 9bc9f224e58553b4cc3e04f7323f0fa01aa69661

PR #89, opened 2022-03-04
Updated OpenGraph image

j0nas:
# Updated OpenGraph image

## Intent

Received an updated image from Sindre, where the dates have been modified.


oyvind-stenhaug (approved):
Looks good! And now we got to make use of the "cache-breaking" functionality :)

PR #94, opened 2022-03-07
Hotfix: render /terms-and-conditions

j0nas:
# Hotfix: render /terms-and-conditions

## Intent

Make /terms-and-conditions render again.

## Details

- "community@sanity.io" link [here](https://stagingadmin.structuredcontent.live/desk/article;8db38690-28be-4d0a-87fb-96f70b850d16) did not have a `href` set, which passed an `undefined` `entity` to _web/util/entityPaths.ts_.
- I fixed the link and published, but the link's href was still `#`; apparently `internalLink`s can have a `value.href`. Let's use this value if present

## Testing this PR

Verify that [/terms-and-conditions](https://structured-content-2022-web-2zq6bgbtz.sanity.build/terms-and-conditions) renders fine without any error/warning output to the console.

### DIFF
--- a/web/components/TextBlock/TextBlock/TextBlock.tsx
+++ b/web/components/TextBlock/TextBlock/TextBlock.tsx
@@ -43,7 +43,7 @@ const components: Partial<PortableTextComponents> = {
     code: ({ children }) => <code>{children}</code>,
     internalLink: ({ text, value }) => (
       /* href has been seen in a case where both link and internalLink marks applied */
-      <Link href={value?.href || getEntityPath(value)}>{text}</Link>
+      <Link href={value?.href || getEntityPath(value.reference)}>{text}</Link>
     ),
   },
 };
--- a/web/components/TextBlock/TextBlock/TextBlock.tsx
+++ b/web/components/TextBlock/TextBlock/TextBlock.tsx
@@ -42,6 +42,7 @@ const components: Partial<PortableTextComponents> = {
     underline: ({ children }) => <u>{children}</u>,
     code: ({ children }) => <code>{children}</code>,
     internalLink: ({ text, value }) => (
+      /* href has been seen in a case where both link and internalLink marks applied */
       <Link href={value?.href || getEntityPath(value)}>{text}</Link>
     ),
   },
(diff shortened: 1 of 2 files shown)


web/components/TextBlock/TextBlock/TextBlock.tsx
```diff
@@ -42,7 +42,7 @@ const components: Partial<PortableTextComponents> = {
     underline: ({ children }) => <u>{children}</u>,
     code: ({ children }) => <code>{children}</code>,
     internalLink: ({ text, value }) => (
-      <Link href={getEntityPath(value.reference)}>{text}</Link>
+      <Link href={value?.href || getEntityPath(value)}>{text}</Link>
```
oyvind-stenhaug, line 45: Nit: The href is OK, but replacing `value.reference` with `value` makes no difference that I can see. Neither of those values is actually an `Entity`. Loading the entire entity for every internal link instead of just its slug also sounds pretty heavy anyway.

Besides, the entire `getEntityPath` helper seems to be outdated ever since https://github.com/sanity-io/structured-content-2022/pull/49 was merged. It handles `session`, `person` and `venue` by creating URLs that are no longer handled. But the possible references in the Studio schema are actually `route`, `person`, `session` and `article`, where at least `route` and `article` might be expected to work at this point.

But this might be considered out of scope for this PR.

oyvind-stenhaug (approved):
It's not actually possible for `internalLink` to have a `href` according to the schema, but it looks like this text is marked both as `link` and as `internalLink` at the same time, so maybe `@portabletext/react` gets confused.

Quickfix approved, but would prefer to have a code comment about it since it looks so out of place.

web/components/TextBlock/TextBlock/TextBlock.tsx
j0nas: Yeah, the property deletion was accidental, reverted in adaf179. I don't really know what to do with `getEntityPath` at this point, so I'll just leave it as-is. I filed #96 for it.

oyvind-stenhaug (approved):
Re-approved. Can you follow up with Celine/Shortcut too maybe? Basically, internal links don't work. (I think it only affects articles at this point.)

PR #95, opened 2022-03-07
TextAndImage.tsx: render Figure's alt text if present

j0nas:
# TextAndImage.tsx: render Figure's alt text if present

## Intent

Fixes [this](https://app.shortcut.com/sanity-io/story/15878/sponsorship-image-alt-text-not-displayed) Shortcut Story.

## Testing this PR

Inspect the "30% Content Strategists ..." image at [/sponsorship-information](https://structured-content-2022-web-cugds69ih.sanity.build/sponsorship-information) and verify that the HTML tag contains an `alt` attribute with the content `30% Content Strategists. 30%  Engineers. 40% Technology Executives.`

### DIFF
--- a/web/components/TextBlock/TextAndImage/TextAndImage.tsx
+++ b/web/components/TextBlock/TextAndImage/TextAndImage.tsx
@@ -24,7 +24,7 @@ export const TextAndImage = ({
           {/* eslint-disable-next-line @next/next/no-img-element */}
           <img
             src={imageUrlFor(image).ignoreImageParams().url()}
-            alt={image.alt}
+            alt={image.alt || ''}
             className={styles.image}
           />
         </div>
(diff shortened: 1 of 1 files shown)


web/components/TextBlock/TextAndImage/TextAndImage.tsx
```diff
@@ -23,7 +24,7 @@ export const TextAndImage = ({
           {/* eslint-disable-next-line @next/next/no-img-element */}
           <img
             src={imageUrlFor(image).ignoreImageParams().url()}
-            alt=""
+            alt={image.alt}
```
oyvind-stenhaug, line 27: Correct me if I'm wrong, but it looks like React's "conditional rendering" means the alt attribute will be completely missing if `image.alt` is undefined, and that this will happen if the field is empty.

Given that there actually is a field for filling in a text alternative, I think we should set an empty attribute if the field is empty, so the editors have the ability to leave it empty for images that are ancillary or purely decorative. Omitting the attribute entirely is only appropriate in a couple of very specific circumstances.

```suggestion
            alt={image.alt || ''}
```

oyvind-stenhaug (changes requested):

web/components/TextBlock/TextAndImage/TextAndImage.tsx
j0nas: Good catch!

oyvind-stenhaug (approved):

PR #100, opened 2022-03-07
Fix supported internal links

j0nas:
# Fix supported internal links

## Intent

- Support only `route` and `article` `internalLink` types; fix `getEntityPath` URL resolution for those two entity types
- Fetch `markDefs` references for articles, which fixes internal links in _/article/:slug_

## Description

I've confirmed with Celine that `route` and `article` are the only two entities we support for `internalLink`. It then makes sense to me to link to articles using the `/article/:slug` path. This is not an editor-only route, I don't know where I got that from. From the related [Story](https://app.shortcut.com/sanity-io/story/15481/routing-update-to-handle-editorial-articles), my impression from @kmelve's wording is that the `/article/:slug` -> `/:slug` rewriting we do is an "extra" and that the "canonical" URL which we should link to is the former. This also matches with the `canonical` attribute we set for article URLs, both rewritten and not.

## Testing this PR

I've created an [Editorial Article](https://stagingadmin.structuredcontent.live/desk/article;72149d3a-1f8c-4be6-822c-ef313370c03d%2Ctemplate%3Darticle) which is available at [/article/test](https://structured-content-2022-web-qvs1gafr3.sanity.build/article/test). Verify that the links on the page point to expected targets.

### DIFF
--- a/web/pages/article/[slug].tsx
+++ b/web/pages/article/[slug].tsx
@@ -20,7 +20,10 @@ const QUERY = groq`
         ...,
         markDefs[] {
           ...,
-          reference->,
+          reference-> {
+            _type,
+            slug,
+          },
         },
       },
     },
--- a/web/pages/article/[slug].tsx
+++ b/web/pages/article/[slug].tsx
@@ -14,7 +14,16 @@ import MetaTags from '../../components/MetaTags';
 
 const QUERY = groq`
   {
-    "article": *[_type == "article" && slug.current == $slug][0],
+    "article": *[_type == "article" && slug.current == $slug][0] {
+      ...,
+      content[] {
+        ...,
+        markDefs[] {
+          ...,
+          reference->,
+        },
+      },
+    },
     "home": *[_id == "${mainEventId}"][0] {
       "ticketsUrl": microcopy[key == "mainCta"][0].action,
     },
--- a/web/util/entityPaths.ts
+++ b/web/util/entityPaths.ts
@@ -1,8 +1,9 @@
 import urlJoin from 'proper-url-join';
+import { Slug } from '../types/Slug';
 
 export const mainEventId = 'aad77280-6394-4090-afad-1c0f2a0416c6';
 
-export const getEntityPath = (entity?: any) => {
+export const getEntityPath = (entity?: { _type: string; slug: Slug }) => {
   if (!entity?.slug?.current) {
     return '#';
   }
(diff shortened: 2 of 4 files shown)


web/pages/article/[slug].tsx
```diff
@@ -14,7 +14,16 @@ import MetaTags from '../../components/MetaTags';
...
 const QUERY = groq`
   {
-    "article": *[_type == "article" && slug.current == $slug][0],
+    "article": *[_type == "article" && slug.current == $slug][0] {
+      ...,
+      content[] {
+        ...,
+        markDefs[] {
+          ...,
+          reference->,
```
Comment (oyvind-stenhaug) [inline web/pages/article/[slug].tsx:23]: When I try this query in Vision it seems to fetch lots more content than needed, e.g. the entire rich text contents of the article in the "Link 2" case. I would think that this means the client-side JSON payload is that much larger if you navigate to the article (that contains the internalLink) via Next.js `<Link>` navigation?

Can this be limited to only the fields we actually need? Which I think would be these:

```suggestion
          reference->{_type, slug},
```

web/util/entityPaths.ts
```diff
@@ -1,23 +1,19 @@
...
-import { Session } from '../types/Session';
-import { Venue } from '../types/Venue';
-import { Person } from '../types/Person';
-
-type Entity = Person | Session | Venue;
 
 export const mainEventId = 'aad77280-6394-4090-afad-1c0f2a0416c6';
 
-export const getEntityPath = (entity?: Entity) => {
+export const getEntityPath = (entity?: any) => {
```
oyvind-stenhaug, line 5: If the query is adjustes per the comment above, then instead of `any`, could this be defined as a simpler object type along the lines of `{_type: string; slug?: Slug;}` or some such?

oyvind-stenhaug (changes requested):

web/pages/article/[slug].tsx
Comment (j0nas) [reply web/pages/article/[slug].tsx]: Good catch! 7c9e547.

web/util/entityPaths.ts
j0nas: 3677fcb makes sense with the comment above. :100:

oyvind-stenhaug (approved):
Looks good to me!

(For links to the Vercel-deployed version for the branch, I like to use the host name that is derived from the branch name when writing the PR description. That way it's easier to re-test things in a newer commit :))

j0nas:
> (For links to the Vercel-deployed version for the branch, I like to use the host name that is derived from the branch name when writing the PR description. That way it's easier to re-test things in a newer commit :))

I forget that that's a thing, thank you for the reminder! Will save me some time updating READMEs.

PR #108, opened 2022-03-10
Update favicons

oyvind-stenhaug:
# Update favicons

## Intent

Update the site icons to new versions provided by Sindre, in Shortcut story "[Minor design implementation tweaks (first round)](https://app.shortcut.com/sanity-io/story/16013/minor-design-implementation-tweaks-first-round)"

## Description

Split this off from the other tweaks since it might be slightly less straightforward, considering cache behavior and such.

As discussed in the story, this cuts back on the number of (more or less proprietary) icon variants/sizes that previously existed, in favor of the sizes Sindre provided in the .zip file (which are served up as standard rel=icon). Also removed a theme color that was set to the old light blue.

I don't have an iPhone, but I've verified that adding the site to the start screen works OK with Chrome on my Android phone. Also tested pinning a tab in Safari (in BrowserStack), and it looks like it picks up a standard icon if the non-standard "mask-icon" has not been specified.

Wilhelm tested adding to the home screen in iOS and it turned out that Safari still won't use standard icons, so I added the biggest version (128x128) as an "apple-touch-icon". This seemed to work fine. Currently it's just a flat color anyway, so a bit of scaling up or down shouldn't be any issue at all.

Also moved the linked icons so that they're getting "cache-busting" URLs and can be served with a long max-age. This was intended as more of a bonus thing, commit can be reverted if the old `_document` approach is preferred.

## Testing this PR

Step-by-step instructions which reviewers can use to verify that this PR
achieves its aim. Specify any prerequisites and assumptions. Be explicit.

1. Open http://structured-content-2022-web-git-update-favicons.sanity.build/ in a browser that shows icons in its tabs, e.g. Firefox or Chromium on Linux
2. Verify that the icon shown in the browser tab is a yellow square with rounded corners, not a Sanity "S"


j0nas (approved):
> Also moved the linked icons so that they're getting "cache-busting" URLs and can be served with a long max-age. This was intended as more of a bonus thing, commit can be reverted if the old _document approach is preferred.

This is a nice little bonus! :100:

PR #109, opened 2022-03-10
Random shapes in NavBlock

j0nas:
# Random shapes in NavBlock

## Intent

Replaces hard-coded shapes in NavBlock with all `FakeItem`s having 33% chance to be a random shape.
Fixes part 1 of [this Shortcut Story](https://app.shortcut.com/sanity-io/story/16075/animate-nav-block).

## Description

(I also took a shot at animating the NavBlock using [react-awesome-reveal](https://github.com/morellodev/react-awesome-reveal), but the library wraps its child elements in `div`s and `span`s, and also seems to not apply the `childClassName` prop to its children. I need to consider other approaches here, and am postponing the implementation to a separate PR so that this simpler code can be shipped.)

The `useEffect` in `FakeItem` is only executed client-side, so this approach avoids the "CSS class names on client and server did not match" error.

## Testing this PR

Visit the [frontpage](https://structured-content-2022-web-bm6g9swt3.sanity.build/) of the Preview Deploy for this PR, and refresh a couple times. Verify that the NavBlock shapes change, and that no errors or warnings are logged to the console.

### DIFF
--- a/web/components/NavBlock/NavBlock.tsx
+++ b/web/components/NavBlock/NavBlock.tsx
@@ -1,33 +1,44 @@
 import Link from 'next/link';
 import clsx from 'clsx';
 import styles from './NavBlock.module.css';
+import { useEffect, useState } from 'react';
+
+const RANDOM_SHAPE_PERCENT_CHANCE = 0.33;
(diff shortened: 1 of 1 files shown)


oyvind-stenhaug (approved):
Looks good to me! One minor suggestion.

PR #111, opened 2022-03-10
External links in new tab / new link structure

oyvind-stenhaug:
# External links in new tab

## Intent

Open rich text "External link"s in a new tab if the "Open in new tab" setting has been switched on, and similarly for richtext "Link" objects and for "Simple call to action" objects.

## Description

This has been discussed in the Shortcut story "[Minor design implementation tweaks (first round)](https://app.shortcut.com/sanity-io/story/16013/minor-design-implementation-tweaks-first-round)", which includes this task ("Open new tab when clickling on external links"). This PR doesn't open _all_ external links in new tabs, just some of them. Mainly, those richtext or `simpleCallToAction` links where the "Open in new tab or window" setting has been enabled.

The PR also currently sets the "Tickets" links in header nav and "nav block" to open in a new tab, because it was requested by Sindre in the story. Would be good to have confirmation from @kmelve if this is OK or if it should be reverted. (As discussed in the story there are also other links that already open in new tabs prior to this PR, these are in the footer and cookie banner.)

## Testing this PR

1. Ensure you're using a browser config with the usual behavior of defaulting to opening links in the current tab (with the possible exception of downloads, non-HTTP/HTTPS URI schemes etc)
2. Open https://structured-content-2022-web-git-external-links-in-new-tab.sanity.build/link-testing
3. Verify that all links/button on the page are working (the ones marked "internal" point to the /programs page, and the ones marked "external" point to sanity.io) and that the ones marked "new tab" open in a new tab (or a new window if your browser is configured like that)

### DIFF
--- a/studio/src/deskStructure.ts
+++ b/studio/src/deskStructure.ts
@@ -169,7 +169,8 @@ export default () =>
             })
         ),
       S.documentTypeListItem("sharedSections").title("Shared Sections"),
-      S.listItem()
+      S.documentTypeListItem("navigation").title("Navigation"),
+      /* S.listItem()
         .title("Navigation")
         .icon(MenuIcon)
         .child(
@@ -195,7 +196,7 @@ export default () =>
               }).icon(MenuIcon),
             ])
             .title("Navigation")
-        ),
+        ), */
       S.documentTypeListItem("article").title("Editorial Articles"),
       S.divider(),
       S.documentTypeListItem("ticket").title("Ticket types"),
--- a/studio/src/deskStructure.ts
+++ b/studio/src/deskStructure.ts
@@ -170,33 +170,6 @@ export default () =>
         ),
       S.documentTypeListItem("sharedSections").title("Shared Sections"),
       S.documentTypeListItem("navigation").title("Navigation"),
-      /* S.listItem()
-        .title("Navigation")
-        .icon(MenuIcon)
-        .child(
-          S.list()
-            .items([
-              createDeskHierarchy({
-                title: "Primary Navigation",
-
-                // The hierarchy will be stored in this document ID 👇
-                documentId: "primary-nav",
-
-                // Document types editors should be able to include in the hierarchy
-                referenceTo: ["route"],
-              }).icon(MenuIcon),
-              createDeskHierarchy({
-                title: "Secondary Navigation",
-
-                // The hierarchy will be stored in this document ID 👇
-                documentId: "secondary-nav",
-
-                // Document types editors should be able to include in the hierarchy
-                referenceTo: ["route"],
-              }).icon(MenuIcon),
-            ])
-            .title("Navigation")
-        ), */
       S.documentTypeListItem("article").title("Editorial Articles"),
       S.divider(),
       S.documentTypeListItem("ticket").title("Ticket types"),
--- a/web/components/Nav/Nav.tsx
+++ b/web/components/Nav/Nav.tsx
@@ -82,11 +82,15 @@ export const Nav = ({ onFrontPage, currentPath, ticketsUrl }: NavProps) => {
           </Link>
           <ul className={clsx(styles.items, !menuOpened && styles.menuClosed)}>
             <li className={styles.ticketItem}>
-              <Link href={ticketsUrl}>
-                <a className={styles.link} onClick={closeMenu}>
-                  Tickets
-                </a>
-              </Link>
+              <a
+                className={styles.link}
+                href={ticketsUrl}
+                target="_blank"
+                rel="noreferrer"
+                onClick={closeMenu}
+              >
+                Tickets
+              </a>
             </li>
             <BasicMenuItem urlPath="/program" label="Program" />
             <BasicMenuItem
@@ -100,7 +104,7 @@ export const Nav = ({ onFrontPage, currentPath, ticketsUrl }: NavProps) => {
             <BasicMenuItem urlPath="/about" label="About" />
           </ul>
           <div className={styles.ticketButton}>
-            <ButtonLink url={ticketsUrl} text="Tickets" />
+            <ButtonLink url={ticketsUrl} text="Tickets" openInNewTab={true} />
           </div>
         </div>
       </GridWrapper>
--- a/web/components/SimpleCallToAction/SimpleCallToAction.tsx
+++ b/web/components/SimpleCallToAction/SimpleCallToAction.tsx
@@ -4,13 +4,19 @@ import urlJoin from 'proper-url-join';
 
 export const SimpleCallToAction = ({ text, link }: SimpleCallToActionProps) => {
   const openInNewTab = Boolean(link?.blank);
-  return text && link?.internal?.slug?.current ? (
-    <ButtonLink
-      text={text}
-      url={urlJoin(link.internal.slug.current)}
-      openInNewTab={openInNewTab}
-    />
-  ) : text && link?.external ? (
-    <ButtonLink text={text} url={link.external} openInNewTab={openInNewTab} />
-  ) : null;
+  if (text && link?.internal?.slug?.current) {
+    return (
+      <ButtonLink
+        text={text}
+        url={urlJoin(link.internal.slug.current)}
+        openInNewTab={openInNewTab}
+      />
+    );
+  }
+  if (text && link?.external) {
+    return (
+      <ButtonLink text={text} url={link.external} openInNewTab={openInNewTab} />
+    );
+  }
+  return null;
 };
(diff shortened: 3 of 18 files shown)


kmelve:
Push a commit that introduces a slight refactor of how links work. This will require migration of data and the frontend as well. 

[Here are the new data structures](http://localhost:3333/desk/sharedSections;eeda6658-b9d0-4abc-adf9-5838222d9aa9%2Cinspect%3Don):
```json
[
  {
    "_key": "f275d1684e71",
    "_type": "articleSection",
    "content": [
      {
        "_key": "2403ca1d576e",
        "_type": "block",
        "children": [
          {
            "_key": "e725b3cc3d0e",
            "_type": "span",
            "marks": ["b2a37f42c659"],
            "text": "Internal Link"
          }
        ],
        "markDefs": [
          {
            "_key": "b2a37f42c659",
            "_type": "link",
            "internal": {
              "_ref": "09489684-de8b-43ce-8888-0c285252bb49",
              "_type": "reference"
            }
          }
        ],
        "style": "normal"
      },
      {
        "_key": "50c948d411b7",
        "_type": "block",
        "children": [
          {
            "_key": "a0d9d4929fb9",
            "_type": "span",
            "marks": ["fde86dd15fdd"],
            "text": "External Link"
          }
        ],
        "markDefs": [
          {
            "_key": "fde86dd15fdd",
            "_type": "link",
            "external": "https://google.com"
          }
        ],
        "style": "normal"
      },
      {
        "_key": "c1163480dac2",
        "_type": "block",
        "children": [
          {
            "_key": "1fcef40bbe40",
            "_type": "span",
            "marks": ["6ea994f9fed0"],
            "text": "Internal Link open in new tab"
          }
        ],
        "markDefs": [
          {
            "_key": "6ea994f9fed0",
            "_type": "link",
            "blank": true,
            "internal": {
              "_ref": "09489684-de8b-43ce-8888-0c285252bb49",
              "_type": "reference"
            }
          }
        ],
        "style": "normal"
      },
      {
        "_key": "e55612b7da60",
        "_type": "block",
        "children": [
          {
            "_key": "e69668cfd2eb",
            "_type": "span",
            "marks": ["f1c19db6344e"],
            "text": "External Link open in new tab"
          }
        ],
        "markDefs": [
          {
            "_key": "f1c19db6344e",
            "_type": "link",
            "blank": true,
            "external": "https://google.com"
          }
        ],
        "style": "normal"
      },
      {
        "_key": "16429da4c8aa",
        "_type": "simpleCallToAction",
        "text": "Internal link",
        "link": {
          "_type": "link",
          "internal": {
            "_type": "reference",
            "_ref": "09489684-de8b-43ce-8888-0c285252bb49"
          }
        }
      },
      {
        "_key": "b768ca3ac499",
        "_type": "simpleCallToAction",
        "text": "External Link",
        "link": {
          "_type": "link",
          "external": "https://google.com"
        }
      },
      {
        "_key": "a90cc94e3c26",
        "_type": "simpleCallToAction",
        "text": "Internal link in new tab",
        "link": {
          "_type": "link",
          "internal": {
            "_type": "reference",
            "_ref": "09489684-de8b-43ce-8888-0c285252bb49"
          },
          "blank": true
        }
      },
      {
        "_key": "1a26bb175355",
        "_type": "simpleCallToAction",
        "text": "External link in new tab",
        "link": {
          "_type": "link",
          "external": "https://www.google.com",
          "blank": true
        }
      }
    ]
  }
]
```

kmelve:
It's not _that_ much data to move, so it seems like it will be quick enough to do a manual transfer. I'll do a migration in the `external-links-in-new-tab` dataset so that we can prepare the frontend.

oyvind-stenhaug:
The current queries are not consistently defining `simpleBlockContent` etc in such a way that the `"_type": "link"` can be fully supported without a lot of GROQ refactoring work, as I see it. https://github.com/sanity-io/structured-content-2022/pull/115 has now been filed for that refactoring.

Once that has been merged, I can update this PR accordingly.

oyvind-stenhaug:
(The aforementioned refactoring PR has been merged.)

studio/src/deskStructure.ts
```diff
@@ -195,7 +196,7 @@ export default () =>
               }).icon(MenuIcon),
             ])
             .title("Navigation")
-        ),
+        ), */
```
j0nas, line 199: Nit: this commented-out code should probably be deleted instead?

web/components/Nav/Nav.tsx
```diff
@@ -100,7 +104,7 @@ export const Nav = ({ onFrontPage, currentPath, ticketsUrl }: NavProps) => {
             <BasicMenuItem urlPath="/about" label="About" />
           </ul>
           <div className={styles.ticketButton}>
-            <ButtonLink url={ticketsUrl} text="Tickets" />
+            <ButtonLink url={ticketsUrl} text="Tickets" openInNewTab={true} />
```
j0nas, line 107:
```suggestion
            <ButtonLink url={ticketsUrl} text="Tickets" openInNewTab />
```

(Nit.)

web/components/SimpleCallToAction/SimpleCallToAction.tsx
```diff
@@ -2,13 +2,15 @@ import { SimpleCallToAction as SimpleCallToActionProps } from '../../types/Simpl
...
+  return text && link?.internal?.slug?.current ? (
+    <ButtonLink
+      text={text}
+      url={urlJoin(link.internal.slug.current)}
+      openInNewTab={openInNewTab}
+    />
+  ) : text && link?.external ? (
+    <ButtonLink text={text} url={link.external} openInNewTab={openInNewTab} />
   ) : null;
+};
```
j0nas, line 16: Nit: is this any easier to read? Or some other variant eliminating the nested ternary.

```suggestion
export const SimpleCallToAction = ({ text, link }: SimpleCallToActionProps) =>
  !text || !link || !(link.internal?.slug?.current || link.external) ? null : (
    <ButtonLink
      text={text}
      openInNewTab={Boolean(link?.blank)}
      url={link.external || urlJoin(link.internal?.slug?.current)}
    />
  );
```

j0nas (approved):
Great job! I also really like the testing page you made, made it super smooth to test this. :100:

web/components/Nav/Nav.tsx
oyvind-stenhaug: Thanks, I sometimes forget some of these "uncanny valley" JSX-vs-HTML things.

I found this note in the ["JSX in depth" docs](https://reactjs.org/docs/jsx-in-depth.html#props-default-to-true), though: "In general, we don’t recommend _not_ passing a value for a prop, because it can be confused with the [ES6 object shorthand](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Operators/Object_initializer#New_notations_in_ECMAScript_2015) `{foo}` which is short for `{foo: foo}` rather than `{foo: true}`. This behavior is just there so that it matches the behavior of HTML."

Still prefer the "value-less" syntax? I don't really have any opinions on it myself.

web/components/SimpleCallToAction/SimpleCallToAction.tsx
oyvind-stenhaug: Personally I find that alternative harder to read.

Would you prefer a sequence of returns, e.g. like this?

```suggestion
export const SimpleCallToAction = ({ text, link }: SimpleCallToActionProps) => {
  const openInNewTab = Boolean(link?.blank);
  if (text && link?.internal?.slug?.current) {
    return (
      <ButtonLink
        text={text}
        url={urlJoin(link.internal.slug.current)}
        openInNewTab={openInNewTab}
      />
    );
  }
  if (text && link?.external) {
    return (
      <ButtonLink text={text} url={link.external} openInNewTab={openInNewTab} />
    );
  }
  return null;
};
```

Another alternative could be to `let url = null;` along with an if/elseif assignment instead of doing a single assignment to a nested/chained ternary expression. I personally don't mind `let`, I just try to avoid them when possible because I got the impression that some other people prefer const.

I anticipate another upcoming variation between internal and external here, by the way, related to the arrow styling. Might affect the readability of the various alternatives one way or the other.

j0nas: I like your "suggested" suggestion, which also seems more malleable wrt. future iterations. :slightly_smiling_face: (But again, this is a nit, feel absolutely free to go for whichever version you prefer.)

web/components/Nav/Nav.tsx
j0nas: TIL! :eyes: I was just pushing my personal preference (hence the nit :stuck_out_tongue:) but if this is _against_ what's recommended, I'll have to change my ways. :sweat_smile:

studio/src/deskStructure.ts
kmelve: Removed!

kmelve (approved):

PR #114, opened 2022-03-15
Nav micro animations

### LINKED ISSUE #23: Front page header doesn't appear if page is loaded at a scrolled-down position
https://github.com/sanity-io/structured-content-2022/pull/22#pullrequestreview-886055231

(happens in Firefox as well, probably not browser-specific)

As the code comment says, ideally this function should probably use IntersectionObserver instead of a scroll listener + hardcoded offset limit (which doesn't take into consideration that the content has different heights with different viewport sizes etc)

j0nas:
# Nav micro animations

## Intent

- Nav slide in/out animation on frontpage
- Menu links slide in/out on small viewports

Fixes [this](https://app.shortcut.com/sanity-io/story/15474/implement-micro-animations) Shortcut Story with input and approval from Sindre.

## Details

(Was told to disregard the "Buttons" item in the associated Story.)

This PR also fixes #23. 

## Testing this PR

1. Open the Preview Deploy for this branch
2. Scroll down to where the Nav header usually would appear and verify that it slides in
3. Scroll back up and verify that it slides out again instead of disappearing
4. Open "Responsive Design Mode" and shrink the viewport until the "Menu" button appears
5. Click the button to open the menu and verify that the items are animated in one-by-one, and that closing and opening the menu again consistently triggers the same animation


web/components/Nav/Nav.tsx
```diff
@@ -15,36 +17,60 @@ interface NavProps {
...
+    />,
+    <MenuItem
+      key={5}
+      {...{ currentPath, closeMenu }}
+      href="/about"
+      label="About"
+    />,
+  ];
+
+  useInterval(
```
oyvind-stenhaug, line 66: Have you considered setting an inline `animation-delay` style instead (with increasing value for each menu item)? And adding `visibility: hidden` / `visibility: visible` to the keyframes (not completely sure if it's sufficient to just use the `from` and `to` keyframes or if there should be an extra keyframe with a really low percentage to ensure that the `visible` part triggers righ away). The `animation-fill-mode` property can be used to ensure the link stays hidden during the delay interval.

If this is possible, then we would
- avoid the extra code from `useInterval`
- avoid mutating the DOM during the animation
- avoid having to re-define the interval length and the media query here

web/pages/app.module.css
```diff
@@ -14,14 +14,38 @@
     position: sticky;
     top: 0;
     left: 0;
+    transform: translateY(-100%);
```
oyvind-stenhaug, line 17: The menu is now missing on non-frontpage pages. (See also below)

web/pages/app.module.css
```diff
@@ -14,14 +14,38 @@
...
+  }
+
+  .slideOut {
+    animation: slideOut 0.5s cubic-bezier(0.25, 0.92, 0.37, 1) forwards;
```
oyvind-stenhaug, line 21: Could all of this be simplified if we use `transition` instead of `animation`? Something like (not tested)

```
.header.onFrontPage {
  position: fixed;
  transform: translateY(-100%);
  transition: translateY 0.5s cubic-bezier(0.25, 0.92, 0.37, 1);
}

.header.onScrolledFrontPage {
  transform: translateY(0%);
}
```

This way I'm thinking there's no need for multiple keyframes/animations, `seenHeader` or `slideOut` class.

oyvind-stenhaug (changes requested):
The header menu's shadow is showing when it's hidden on the front page. Is that expected?

web/components/Nav/Nav.tsx
oyvind-stenhaug: Taking a cue from the navblock animation PR, the inline style could alternatively be a custom property (e.g. `--delay`), for better decoupling. Presumably pretty straightforward to set up in such a way that it falls back gracefully (i.e. just shows the menu items without animating) in IE11 which would drop the CSS declaration(s) that use that property.

web/components/Nav/Nav.tsx
```diff
@@ -15,36 +17,60 @@ interface NavProps {
 export const Nav = ({ onFrontPage, currentPath, ticketsUrl }: NavProps) => {
   const [menuOpened, setMenuOpened] = useState(false);
   const contentsId = 'nav-menu-contents';
+  const [menuItemsCount, setMenuItemsCount] = useState(0);
 
   useEffect(() => {
     document.body.classList.toggle('main-menu-open', menuOpened);
+    if (!menuOpened) {
+      setMenuItemsCount(0);
```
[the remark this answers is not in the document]
j0nas: Sure. :slightly_smiling_face: c206434

web/components/Nav/Nav.tsx
j0nas: This is a _great_ suggestion, thank you. 28e906b.

web/pages/app.module.css
j0nas: Another great suggestion! bc6daef 

(~Note that this re-introduces the issue of the header not appearing of the page is loaded with a non-zero scrollTop position.~ Also fixed in 2fd8be9.)

j0nas:
> The header menu's shadow is showing when it's hidden on the front page. Is that expected?

@oyvind-stenhaug Not initially.. but I really like how it looks. Can we keep it? :smile: 

I also caught a large performance sink in 861c1fb.

web/pages/app.module.css
```diff
@@ -18,10 +18,12 @@
...
   .header.onFrontPage {
     position: fixed;
-    display: none;
+    transform: translateY(-100%);
+    transition: transform 0.5s cubic-bezier(0.25, 0.92, 0.37, 1);
   }
 
   .header.onScrolledFrontPage {
-    display: block;
+    position: fixed;
```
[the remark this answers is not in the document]
j0nas: Yeah, makes sense. b8c8182

j0nas:
@oyvind-stenhaug 

> [...] I noticed that the Figma sketch has a comment from Sindre about the mobile menu slide animation that specifies a different behavior (250ms animation duration, not 0.5s; 100ms delay increments, not 50ms; 20px offset, not 75%). Don't know if the difference here is intentional, I don't see any comments in the Shortcut story about it.

Yes, these were arrived at in collaboration with Sindre, I should have been explicit about that in the PR readme. I thought I noted that down in the commit descriptions (hence the more-than-usually erratic commits), but I apparently did not. :sweat: 

> Also, the same Figma comment specifies an animation on the background color of the mobile menu. I remember Sindre mentioned this when he showed me the sketch, before we redistributed the implementation tasks between us. As I recall, the color would be animated from transparent (orange) to orange.

Yep, Sindre also opted to drop that in the end. Thank you for double-checking on this, I'll try my best to be more explicit about these things in the future. :+1: 

> One benefit would be that we'd be able to specify multiple (perhaps completely unrelated) animations on the elements, without getting the same delay everywhere (or having to resort to !important). So far just theoretical, of course.

I filed #130 for this.

> That's really a design question. The Figma animation mockup doesn't have such a shadow in the hidden state, but I'm not going to let it trigger a "blocking" review. I just wanted to ensure it was intentional.

I asked Sindre about this and he didn't like it :crying_cat_face: Hid it in b58c231.

PR #115, opened 2022-03-15
Refactor and fix GROQ queries

oyvind-stenhaug:
# Refactor and fix GROQ queries

## Intent

Rework the GROQ queries, particularly the one for `page` documents with their sections array, with projections so that things like `simpleBlockContent` are defined explicitly in one place, for consistent support of things like internal links and CTAs that are given by references (which need to be expanded in the query).

Also to try to limit the amount of JSON data we load on the client, by using projections to get only what we actually need, as described in a [previously mentioned blog post](https://www.sanity.io/blog/3-simple-things-in-groq-to-supercharge-your-frontend-development#db26fcd247c4).

See Shortcut story ["Refactor GROQ queries and fix external links"](https://app.shortcut.com/sanity-io/story/16228/refactor-groq-queries-and-fix-internal-links)

## Description

As the Shortcut story indicates, this is related to (and is regarded as blocking for) https://github.com/sanity-io/structured-content-2022/pull/111. That PR/branch includes some changes to the Sanity content model's treatment of links, which necessitated updating the parts of the query/-ies that relate to (simple) block content fields. It turned out these fields weren't always specified, which didn't seem like a straightforward thing to resolve. Hence I split it out as a separate story/task/PR, to limit scope.

As a further scope-limiting measure, I've mostly focused on sections that are in active use and that involve block content or call to action fields. E.g. I have not updated the projection for Sponsors/Sponsorships sections. The (shared) sections testing pages already seem to trigger errors and warnings for some of these even on staging.

## Testing this PR

1. Open the [/sections-testing](https://structured-content-2022-web-git-refactor-and-fix-groq-queries.sanity.build/sections-testing) page
2. Scroll down to the Q&A section ("First question" etc) and verify that the "CTA within Q&A →" link shows up (looks like a button)
3. Optionally, check other pages as well and verify that no sections/content is missing

Note that the Q&A heading now also shows up. This is visible both on the /sections-testing page and on [/shared-sections](https://structured-content-2022-web-git-refactor-and-fix-groq-queries.sanity.build/shared-sections)

### DIFF
--- a/web/pages/[[...slug]].tsx
+++ b/web/pages/[[...slug]].tsx
@@ -30,7 +30,7 @@ import {
 import styles from './app.module.css';
 
 const SHARED_SECTIONS = `
-  ...,
+  _type,
   _type == "figure" => { ${FIGURE} },
   _type == "articleSection" => { ${ARTICLE_SECTION} },
   _type == "textAndImageSection" => { ${TEXT_AND_IMAGE_SECTION} },
@@ -67,6 +67,7 @@ const SHARED_SECTIONS = `
     tickets[]->{ ${TICKET} },
     "allTickets": *[_id == "${mainEventId}"][0].tickets[]->{ ${TICKET} }
   },
+  _type == "formSection" => { ... },
   _type == "programsSection" => {
     type,
     programs[]-> { ${PROGRAM} },
--- a/web/pages/[[...slug]].tsx
+++ b/web/pages/[[...slug]].tsx
@@ -19,8 +19,11 @@ import { Section } from '../types/Section';
 import { Hero as HeroProps } from '../types/Hero';
 import { mainEventId } from '../util/entityPaths';
 import {
+  ARTICLE_SECTION,
   FIGURE,
   HERO,
+  QUESTION_AND_ANSWER_COLLECTION_SECTION,
+  TEXT_AND_IMAGE_SECTION,
 } from '../util/queries';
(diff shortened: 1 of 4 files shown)


web/pages/[[...slug]].tsx
```diff
@@ -18,107 +18,75 @@ import { Slug } from '../types/Slug';
...
+  FIGURE,
+  HERO,
+  PROGRAM,
+  QUESTION_AND_ANSWER_COLLECTION_SECTION,
+  TEXT_AND_IMAGE_SECTION,
+  TICKET,
+} from '../util/queries';
 import styles from './app.module.css';
 
+const SHARED_SECTIONS = `
```
Comment (j0nas) [inline web/pages/[[...slug]].tsx:32]: Is this query specific to <em>[[...slug]].tsx</em> or could it also be moved to _../util/queries_?

j0nas (approved):
This is amazing. Fantastically well done! :star_struck: This new structure is super clean and far more scalable.

web/pages/[[...slug]].tsx
Comment (oyvind-stenhaug) [reply web/pages/[[...slug]].tsx:32]: I wasn't sure what was best there. So far it's specific to [[...slug]].tsx though so are many of the others, so it's a gray area. Let's move it if more document types appear later on that also use `sections`, maybe.

(If it's moved to util/queries we'd have to decide if it should take event ID as a parameter or if that file should fetch `mainEventId` itself, I suppose.)

PR #116, opened 2022-03-16
Navblock animation

jorgenblindheim:
(And set state only when needed on scroll)


web/components/NavBlock/NavBlock.module.css
```diff
@@ -2,6 +2,21 @@
   margin: 80px 0;
 }
 
+.navEnter .item,
```
oyvind-stenhaug, line 5: Nit: I think it would be slightly clearer if the "base" class of the nav is also used, to help show where it comes from. (Drawing a parallel to `.block--modifier` in BEM.) I.e. `.nav.navEnter .item`, or maybe better with a class rename, something like `.nav.isIntersecting .item`

Also, could these declaration blocks be moved below the existing `.item, .fakeItem` block? They're roughly ordered by which elements are targeted, in the same order the elements (first) appear in the DOM/JSX, with the "base" styling first and variations after. First for base styling, then again within each `@media` block from narrowest to widest. (Sorry this convention is undocumented.)

web/components/VenueNames/VenueNames.module.css
```diff
@@ -4,6 +4,20 @@
   padding: 0;
 }
 
+.venuesEnter .venue {
```
oyvind-stenhaug, line 7: Nit: equivalent comments to the above on `.navEnter .item`, regarding position and class names. (I.e. I'd move this below the `.venue` ruleset, and maybe do something like `.venues.isIntersecting .venue` instead. Looks like the `.venue:nth-child(n + 2)::before` ruleset is already breaking the convention…)

oyvind-stenhaug (changes requested):
Looks pretty good!

I've mostly focused on the CSS/markup part, since my React skills aren't as strong. Looked through all of it, though. Mainly I just have a question about some warnings (comment inline).

Because the animation triggers as the wrapper element enters the viewport, you won't necessarily see a lot of it on smaller screens. But maybe that's OK?

(E.g. on a 320x568 screen ("iPhone 5/SE iOS 10.3.1" in Firefox' Responsive Design Mode), loading the front page I only see the top half of the three topmost NavBlock items. So I see them sliding/fading into view, but then no more animation as I scroll further down to reveal the other 9 items.)

oyvind-stenhaug:
Is this PR actually a work in progress? If so, maybe use the "draft" state in GitHub, to signal when it's ready for review.

web/components/NavBlock/NavBlock.module.css
jorgenblindheim: Why aren't we using BEM anyway? Then it wouldn't have a specificity war going on in the first place.

Also, if we where to use SASS I think it would be much easier to attach @media blocks, :hovers, etc to the actual classnames making the code much clearer and easier to read.

(But I'll update the classnames so it matches the rest of the css).

web/components/NavBlock/NavBlock.tsx
```diff
@@ -19,15 +22,18 @@ const getRandomShape = (): Shape => {
 };
 
 const FakeItem = ({ divider, mobile, tablet, desktop }: FakeItemProps) => {
-  const [shapeClass, setShapeClass] = useState<string>(null);
-  useEffect(() => {
+  const shapeClass: string | null = useMemo(() => {
     if (Math.random() <= RANDOM_SHAPE_PERCENT_CHANCE) {
```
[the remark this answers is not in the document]
jorgenblindheim: The problem with using useEffect is that the component will render twice which isn't necessary. Not really sure how to debug this though. I have a theory on how to solve it, but where do we see these errors?

web/components/NavBlock/NavBlock.module.css
oyvind-stenhaug: There is no specificity war, just a minor readability nit.

We aren't using BEM because `css-modules` already does the same "namespacing" job, IMO.

(To be clear, any comment we prefix "Nit:" is not regarded as "blocking" the PR approval. Maybe that was already obvious, I don't know how widespread the practice is.)

web/components/NavBlock/NavBlock.tsx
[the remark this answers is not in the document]
oyvind-stenhaug: I see them in the browser console.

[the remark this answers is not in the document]
j0nas: e2e76ae fixes these warnings. The "double-render" method is the idiomatic way of handling this for React/Next.js.

web/components/ConferenceUpdatesForm/ConferenceUpdatesForm.module.css
```diff
@@ -107,3 +116,10 @@
     margin: 0;
   }
 }
+
+@keyframes item-enter {
+  from {
+    opacity: 0;
+    transform: translateY(var(--distance)) rotate(var(--rotation));
+  }
+}
```
j0nas, line 125: Should this animation perhaps be pulled up into _globals.css_ rather than duplicated many places?

j0nas:
@oyvind-stenhaug I've tweaked the animation properties with help from Sindre, and he's happy with the look and feel we've arrived at. This PR's now all yours for code tweaks whenever you have the time. :slightly_smiling_face:

oyvind-stenhaug (approved):
@j0nas I pushed one minor coding style tweak, plus a reversion of a mysterious `--grid-width` change as discussed in Slack. Other than that I read through the code and it looks good to me.

I'm wondering if we should wrap all of these animations inside `@media (prefers-reduced-motion: no-preference) { }`, though. What do you think? (If so, want to do that change while I look at the editorial article PR?)

PR #117, opened 2022-03-16
Editorial Articles: render `relatedTo` entities, `_createdAt`, `_updatedAt`

j0nas:
# Editorial Articles: render `relatedTo` entities, `_createdAt`, `_updatedAt`

## Intent

Renders `relatedTo` (people, sessions, venues) and `authors` entities in Editorial Article pages.

Fixes [this](https://app.shortcut.com/sanity-io/story/16417/editorial-article-front-end) Shortcut Story.

## Description

- Conditionally hiding the pipe (`|`) separating the `_createdAt` and `_updatedAt` wasn't solved in the smoothest way. If you can think of a better solution, feel free to overwrite the current solution, @oyvind-stenhaug.
- The author/person thumbnails get completely blurry if we set the `size` to match that of the Figma sketches -- setting them to 4x of the expected dimensions fixes this blur :shrug: 
- @kmelve mentioned in today's sync that a separate "updated at"-field will be pushed. A separate follow-up PR can handle this minor change if this PR gets approved prior to that being in place.
- @oyvind-stenhaug I'm going to let you work your magic wrt. to the layout changes required to have the positioning of the `relatedTo` entities in large viewports match the Figma sketches if you're OK with that. (Please let me know if you'd rather have me fix it. I don't mind at all, just assuming it'll be faster for you to do it. :slightly_smiling_face:)

## Testing this PR

1. Open this PR's Preview Deploy and navigate to /scholarships
2. Verify that both `authors` and `relatedTo` entities are rendered in accordance with the Figma specifications
    * NB! We're using the `staging` dataset now; I've added these associations for the Scholarships Editorial Article there and not in production :slightly_smiling_face:


oyvind-stenhaug:
> Conditionally hiding the pipe (|) separating the _createdAt and _updatedAt wasn't solved in the smoothest way. If you can think of a better solution, feel free to overwrite the current solution, @oyvind-stenhaug.

Yeah this didn't really work for me (I would see the "|" character even though the text wrapped to two lines, in some viewport sizes) and the lengths of the month/date strings can vary quite a bit, so there is no one "correct" pixel value.

It's probably possible to use some tricks with a background image to do what I'm guessing you were going for, but can we keep it simple and just use the existing `@media` breakpoints? [4ab8102](https://github.com/sanity-io/structured-content-2022/pull/117/commits/4ab8102df2622940d5960fc6df8a608dc5724ef3)

oyvind-stenhaug:
Pushed some more markup/styling adjustments including the `relatedTo` layout. Would also like to use list markup for the "related to" sections (like I've done for authors), and haven't had time to review the rest of this PR either, but running out of time for today.

One thing I noticed was the date formatting, is UTC the one we want there or should it be e.g. `America/Los_Angeles` as a sort of "default"? Minor nit, maybe nobody will notice if it's off by one compared to their implied expectations.

oyvind-stenhaug (changes requested):
@j0nas I went through and updated the HTML/CSS. Looks OK to me now, have asked if Sindre wants to check the rendering as well.

Two small other things:
- A suggestion about the image URL, see inline comment
- Looks like there exist fields for manually specifying created/updated time now, want to add the support for them?

web/components/Card/Card.tsx
```diff
@@ -0,0 +1,28 @@
...
+}
+
+export const Card = ({ children, figure }: CardProps) => (
+  <div className={styles.card}>
+    {figure /* eslint-disable-next-line @next/next/no-img-element */ ? (
+      <img
+        src={imageUrlFor(figure)
+          .size(112, 112)
+          .fit('min')
+          .ignoreImageParams()
```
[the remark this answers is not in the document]
j0nas: Good catch! b8dac10

j0nas:
@oyvind-stenhaug Thank you for looking over this! 

- 3057fc6 renders the new `publishedAt`/`updatedAt` fields
- 7e607c8 makes card images black&white per Sindre's request in the Story
- 0a99647 makes `<aside>` content span four columns instead of five, also per Sindre's requeste

web/pages/article/article.module.css
```diff
@@ -1,20 +1,112 @@
...
+  .extraHeroContent {
+    margin-top: 64px;
+  }
+
+  .publishedAt,
+  .updatedAt {
+    display: inline;
+  }
+
+  .updatedAt::before {
```
oyvind-stenhaug, line 63: After commit 3057fc620877bbb43aa7a2492780f3f865237fb0, it is no longer guaranteed that there is a publishedAt.

Suggestion (not tested):

```suggestion
  .publishedAt + .updatedAt::before {
```

oyvind-stenhaug (changes requested):
Looks OK, except it looks like it's possible to set "Updated at" and omit "Publish time", and that the vertical bar will show up in this case, which seems a bit strange.

web/pages/article/article.module.css
j0nas: Tested all permutations locally, it worked well. Thank you!

oyvind-stenhaug (approved):

PR #118, opened 2022-03-16
add updated at field

kmelve:
# Add updated at field

## Intent

Add a new field to set the updated at time for editorial articles.


j0nas (approved):
Thank you!

PR #120, opened 2022-03-17
add more advanced benefits lists for sponsorships

kmelve:
# Add more advanced benefits lists for sponsorships

## Intent

Express unique measures on shared benefits across sponsorship packages.

## Description

This introduces a new list and a new document type for benefits. The document type is not yet added to the overall structure but is available through the new array field. The thinking is that the entry of a benefit will make its boolean status `true`. If you add a number, then it will override the boolean status. It's an open question if 0 should say 0 or make it `false`. 

This introduces a data model change and content needs to migrated.

### DIFF
--- a/studio/schemas/documents/benefit.ts
+++ b/studio/schemas/documents/benefit.ts
@@ -1,0 +1,12 @@
+export default {
+  name: "benefit",
+  title: "Sponsporship benefit",
+  type: "document",
+  fields: [
+    {
+      name: "name",
+      title: "Name",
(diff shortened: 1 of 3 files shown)


j0nas (approved):
Thanks, this is everything I need for this feature. :100: 

(I wish I'd seen this a bit earlier, but it shouldn't be too much work to adapt what I currently have into using this. Another benefit of doing all aspects of a feature in the same Story and branch, I suppose.)

PR #126, opened 2022-03-21
New link structure for articles

oyvind-stenhaug:
# New link structure for articles

Note: Should not be merged without coordinating the corresponding migration of the live content. This PR is not "backwards compatible" with the old link structure.

## Intent

Update the app code according to the new Sanity schema for links in `blockContent`, which is a rich text type that's primarily used in article content. See also Shortcut story [New links structure for articles](https://app.shortcut.com/sanity-io/story/16694/new-links-structure-for-articles)

## Description

This corresponds to a [commit that was pushed to PR 111](https://github.com/sanity-io/structured-content-2022/pull/111/commits/95a8e56b7412a174240d53922e117a37329d098b). Ideally we would have followed up on that same PR, but I've been busy on other projects myself, so the PR ended up being merged without corresponding changes to `/web`.

I've also updated the `simpleCallToAction` support for `blockContent` fields, which is somewhat related (it also uses the new link structure). It seems slightly odd to me that this is embedded differently in `blockContent` vs `simpleBlockContent`—the former allows it as inline content _within_ a block, while the latter has it as a possible object to mix _in between_ blocks. Is this intentional, @kmelve ?

I also changed one more thing about `SimpleCallToAction`: previously, this was using URLs of the form `/slug`, relying on the rewrites for articles. This seemed inconsistent with the existing (old) `internalLink` support, which used `getEntityPath`. Now this helper is used in both places, which also seems in line with the "/article/slug" URL being tagged as the "canonical" one.

## Testing this PR

For comparison purposes, I've left the existing `staging` content (such as /article/test) alone and created a new article with examples for the new schema.

1. Open https://structured-content-2022-web-git-new-link-structure-for-articles.sanity.build/article/test-links
2. Verify that all the links in the article body are working:
    - "External link" (sanity.io in new tab/window)
    - "Internal link" (Program page in same tab/window)
    - "Program →" (Program page in same tab/window)
    - "Test" →" (Test article in same tab/window)

### DIFF
--- a/web/components/SimpleCallToAction/SimpleCallToAction.tsx
+++ b/web/components/SimpleCallToAction/SimpleCallToAction.tsx
@@ -1,6 +1,6 @@
 import { SimpleCallToAction as SimpleCallToActionProps } from '../../types/SimpleCallToAction';
 import ButtonLink from '../ButtonLink';
-import urlJoin from 'proper-url-join';
+import { getEntityPath } from '../../util/entityPaths';
 
 export const SimpleCallToAction = ({ text, link }: SimpleCallToActionProps) => {
   const openInNewTab = Boolean(link?.blank);
@@ -8,7 +8,10 @@ export const SimpleCallToAction = ({ text, link }: SimpleCallToActionProps) => {
     return (
       <ButtonLink
         text={text}
-        url={urlJoin(link.internal.slug.current)}
+        url={getEntityPath({
+          _type: link.internal._type,
+          slug: link.internal.slug,
+        })}
         openInNewTab={openInNewTab}
       />
     );
--- a/web/components/SimpleCallToAction/SimpleCallToAction.tsx
+++ b/web/components/SimpleCallToAction/SimpleCallToAction.tsx
@@ -8,10 +8,7 @@ export const SimpleCallToAction = ({ text, link }: SimpleCallToActionProps) => {
     return (
       <ButtonLink
         text={text}
-        url={getEntityPath({
-          _type: link.internal._type,
-          slug: link.internal.slug,
-        })}
+        url={getEntityPath(link.internal)}
         openInNewTab={openInNewTab}
       />
     );
--- a/web/components/TextBlock/TextBlock/TextBlock.tsx
+++ b/web/components/TextBlock/TextBlock/TextBlock.tsx
@@ -45,14 +45,11 @@ const components: Partial<PortableTextComponents> = {
       /* href has been seen in a case where both link and internalLink marks applied */
       <Link href={value?.href || getEntityPath(value.reference)}>{text}</Link>
     ),
-    /* There are two different link marks, but currently TextBlock is used for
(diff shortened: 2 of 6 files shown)


web/components/TextBlock/TextBlock/TextBlock.tsx
```diff
@@ -41,18 +41,11 @@ const components: Partial<PortableTextComponents> = {
...
-     * simpleBlockContent (internal/external). Would prefer to refactor the
-     * TextBlock approach, but it'd be a pretty big job.
-     */
     link: ({ children, value }) => {
-      const url =
-        value?.href || value?.internal?.slug?.current || value?.external;
+      const resolvedSlug = value?.internal?.slug?.current
+        ? getEntityPath(value.internal)
+        : null;
+      const url = resolvedSlug || value?.external;
```
j0nas, line 48: Nit: Maybe making `getEnittyPath` return `#` if `slug.current` is falsy was a mistake. :thinking: Very optional: rewrite `getEntityPath` to return something falsy for that case (Empty string? `null`? Not sure what's best), which would allow the following:

```suggestion
      const url = getEntityPath(value.internal) || value?.external;
```

(Not necessarily only to save these few bytes, but moreso because `#` isn't really a sensible default and using `getEntityPath` this way should be possible. But you might not want to scope this whole ordeal within the realm of this PR. Very, very optional comment.)

j0nas (approved):
Nice. All testing steps yield expected results. :heavy_check_mark:

web/components/SimpleCallToAction/SimpleCallToAction.tsx
```diff
@@ -1,14 +1,17 @@
...
   const openInNewTab = Boolean(link?.blank);
   if (text && link?.internal?.slug?.current) {
     return (
       <ButtonLink
         text={text}
-        url={urlJoin(link.internal.slug.current)}
+        url={getEntityPath({
+          _type: link.internal._type,
+          slug: link.internal.slug,
+        })}
```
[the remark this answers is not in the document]
oyvind-stenhaug: Yeah, good catch about `getEntityPath` already handling a null value there despite the type signature. 2908266557a2796efe3211404517a375036a1edf

web/components/TextBlock/TextBlock/TextBlock.tsx
oyvind-stenhaug: Agreed. In order to limit the scope, I went with an empty string. This seems to be safe to pass to `next/link`. Using `null` feels more correct to me, but it's just a feeling and it would require checking that all other call sites could handle it (`next/link` errors out if you feed it `null` for `href`, it seems), so I just added a comment about it. 108c6cbe6bd082b77ca26924ba5fe316df25c76c

j0nas (approved):
Great! Thank you :100:

kmelve:
Not intentional! Pushed a fix for it on this branch,

j0nas (approved):

PR #129, opened 2022-03-22
[[...slug]].tsx: fetch registrationUrl from main event

j0nas:
# [[...slug]].tsx: fetch registrationUrl from main event

## Intent

Fetch `registrationUrl` correctly for _[[...slug]].tsx_ too, unlike #128 which only does so for editorial articles.

## Description

I've deleted the `staging` dataset and copied `production` into `staging` anew. 

## Testing this PR

Verify that the "Tickets" links in the Header and Nav work again in the PR's Preview Deploy.

### DIFF
--- a/web/components/Nav/Nav.tsx
+++ b/web/components/Nav/Nav.tsx
@@ -97,10 +97,7 @@ export const Nav = ({ onFrontPage, currentPath, ticketsUrl }: NavProps) => {
               urlPath="/sponsorship-information"
               label="Sponsorship"
             />
-            <BasicMenuItem
-              urlPath="/registration-info"
-              label="Registration"
-            />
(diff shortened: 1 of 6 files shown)


oyvind-stenhaug (approved):
Approved, but I also note that there is one more usage of `microcopy`, in next.config.js. A redirect that's set up for `/tickets`.

Seems much less urgent to fix than the `[[...slug]]` query, though, hence the approval.

j0nas:
> Approved, but I also note that there is one more usage of `microcopy`, in next.config.js. A redirect that's set up for `/tickets`.

Good catch. Fixed in 6d297f9. :)

oyvind-stenhaug (approved):
LGTM

PR #122, opened 2022-03-18
Render Sponsorships section

j0nas:
# Render Sponsorships section

## Intent

Render the "Sponsorship Packages" table.

## Description

This PR implements table/lists under the "Sponsorship Packages" heading in the Figma sketches, following Studio changes in #120. This PR also includes a bit of refactoring to pull out and reuse some components already defined by _Tickets.tsx_.

## Testing this PR

1. Go to [/sponsorship-information](https://structured-content-2022-web-git-feature-sc-14927sponsors-08bb9f.sanity.build/sponsorship-information)
2. Verify that the table under "Sponsorship Packages" renders all the fields specified by the [Figma sketch](https://www.figma.com/file/XjAvJKsLpxYJSc8QDWK81j/Sanity-SCC-Website-(Shared)?node-id=1758%3A8699) and looks good in all specified viewports
3. Go to [/registration-info](https://structured-content-2022-web-git-feature-sc-14927sponsors-08bb9f.sanity.build/registration-info)
4. Verify that the fields that still render look OK and that the _/Tickets_ component refactorings in this PR don't break the component
    * Note that #119 introduces new fields to Tickets, which likely is why some fields aren't rendered locally/in Preview Deploys
5. (Push desired fixes directly to this PR if you want to)

PR #132, opened 2022-03-24
MetaTags.tsx: absolute path for fallback OG image

j0nas:
# MetaTags.tsx: absolute path for fallback OG image

## Intent

FIxes [this](https://app.shortcut.com/sanity-io/story/16921/default-seo-image-url-isn-t-picked-up) Shortcut Story.

## Testing this PR

1. Go to https://www.bannerbear.com/tools/twitter-card-preview-tool/
2. Enter the Preview URL for this PR into the input field at the top and prees the "Preview" button
3. Verify that the fallback image appears in the preview card
4. (optional) Verify that the fallback image does not appear when testing current [production site](https://structuredcontent.live)

### DIFF
--- a/web/components/MetaTags/MetaTags.tsx
+++ b/web/components/MetaTags/MetaTags.tsx
@@ -3,6 +3,7 @@ import urlJoin from 'proper-url-join';
 import opengraphImage from '../../images/opengraph-image.png';
 import { imageUrlFor } from '../../lib/sanity';
 import { Figure } from '../../types/Figure';
+import { productionUrl } from '../../util/constants';
 
 interface MetaTagsProps {
   title: string;
@@ -33,7 +34,7 @@ export const MetaTags = ({
     <NextSeo
       title={title}
       description={description}
-      canonical={urlJoin('https://structuredcontent.live', canonicalPath)}
+      canonical={urlJoin(productionUrl, canonicalPath)}
       noindex={noIndex}
       openGraph={{
         images: [
@@ -46,7 +47,7 @@ export const MetaTags = ({
                 alt: image.alt,
               }
             : {
-                url: opengraphImage.src,
+                url: urlJoin(productionUrl, opengraphImage.src),
                 width: opengraphImage.width,
                 height: opengraphImage.height,
               },
--- a/web/util/constants.js
+++ b/web/util/constants.js
@@ -1,4 +1,5 @@
 // This file needs to be CJS due to it being referenced by next.config.js.
 module.exports = {
   mainEventId: 'aad77280-6394-4090-afad-1c0f2a0416c6',
+  productionUrl: 'https://structuredcontent.live',
 };


oyvind-stenhaug (approved):
Approved because it solves the issue, but rather than `productionUrl` it would be more useful to have a `baseUrl` or some such, that is different depending on the environment. Probably tricky to do, though, since there is no single staging environment (that we're actually using) and instead the PRs/commits each have their own domain.

Speaking of own domains: I prefer the old style of branch names. With the ones used by this PR and of https://github.com/sanity-io/structured-content-2022/pull/122 it's both
- more difficult to switch between them/use auto completion in `git`, and
- the branch name gets so long that the most useful parts are chopped off in the domain (good: `https://structured-content-2022-studio-git-nav-micro-animations.sanity.build/`; bad: `https://structured-content-2022-studio-git-bug-sc-16921default-s-b2c87d.sanity.build/`)

This is subjective, but I don't personally find the classification of bug vs feature to be useful in the branch name. And if I were to need to find the Shortcut story of an old branch for some reason, I'd get there either via a commit message or simply by looking up the branch in GitHub and getting to both the PR and (through that) the Shortcut story that way.

j0nas:
@oyvind-stenhaug 

> Approved because it solves the issue, but rather than `productionUrl` it would be more useful to have a `baseUrl` or some such, that is different depending on the environment. Probably tricky to do, though, since there is no single staging environment (that we're actually using) and instead the PRs/commits each have their own domain.

Yeah, something like this would be useful to have, but like your comment states, I don't see how we could make this work for staging without resorting to hacks. Feel free to file an issue if you can make this actionable. :slightly_smiling_face: 

> Speaking of own domains: I prefer the old style of branch names. 

The branch name isn't something I personally chose, but rather the result of using Shortcut's expected feature branch naming, [as commented on by Knut in Slack](https://wja.slack.com/archives/C02UKELNTE0/p1647477598895309). Whether we want to stick to this convention or not can be discussed, I merely defaulted to adopting it after it was pointed out to me. In my eyes, the automatic linking to relevant commits/PRs in Shortcut Stories is quite worth it if the tradeoff is merely less pretty preview links. 

> more difficult to switch between them/use auto completion in git, and

Consider looking into the [GitHub CLI](https://cli.github.com/). It has many nifty features, but I only use `gh co #<PR ID>` (having previously ran `gh alias 'co' 'pr checkout'`) which ends up saving me some time and cognitive burden. This is also why you see me preferring to refer in PRs by their IDs on Slack.

oyvind-stenhaug:
> the result of using Shortcut's expected feature branch naming
> In my eyes, the automatic linking to relevant commits/PRs in Shortcut Stories is quite worth it

Aha, I see. In my eyes it's not worth it, I would personally prefer to link them manually (which I've also done in the past) :)

> `gh co #<PR ID>`

Seems our opinion differs on this one too :) I don't remember PR IDs, this wouldn't help me most of the time (on the other hand with directly descriptive branch names I'm often able to switch without having to go look up stuff in the browser).

PR #133, opened 2022-03-25
Speakers page

j0nas:
# Speakers page

## Intent

Page for individual Speakers that have at least one incoming reference, e.g. [Carrie Hane][carrie].
Solves [this](https://app.shortcut.com/sanity-io/story/14187/speaker-details-front-end) Shortcut Story.

## Testing this PR

1. Open /speakers/carrie-hane either locally or in the [latest Preview Deploy][carrie]
2. Verify that the page renders all fields correctly and appears in accordance with [the corresponding Figma sketches](https://www.figma.com/file/XjAvJKsLpxYJSc8QDWK81j/Sanity-SCC-Website-(Shared)?node-id=4476%3A15890) for all supported viewports

[carrie]: https://structured-content-2022-web-git-feature-sc-14187speaker-88f7a5.sanity.build/speakers/carrie-hane


web/hooks/useRandomShape.ts
```diff
@@ -0,0 +1,19 @@
...
+export const useRandomShape = (randomShapeChance: number = 1) => {
+  const [shapeClass, setShapeClass] = useState<string>();
+
+  useEffect(() => {
+    if (!shapeClass && Math.random() <= randomShapeChance) {
+      setShapeClass(styles[`shape${getRandomShape()}`]);
+    }
+  }, [shapeClass, randomShapeChance]);
+
+  return shapeClass;
```
oyvind-stenhaug, line 18: Nit: Is is possible to make this return a string (or null) instead in such a way that all the `.shape` styles are collected in one place instead of being split across two separate files (one in the hooks directory and one in the usual/expected location in the components directory)?

oyvind-stenhaug (changes requested):
I've pushed some commits here… mostly focused on the HTML/CSS. Have also left some comments and questions, though not 100% sure if the "backend" stuff was regarded as finished.

web/pages/speakers/[slug].tsx
```diff
@@ -0,0 +1,213 @@
...
+      index={0}
+      isInline={false}
+      renderNode={null}
+    />
+    <Footer links={footer.links} />
+  </>
+);
+
+export const getStaticPaths: GetStaticPaths = async () => {
+  const allSlugsQuery = groq`*[defined(slug.current) && _type == 'person'][].slug.current`;
```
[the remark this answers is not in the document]
Comment (j0nas) [reply web/pages/speakers/[slug].tsx:170]: I've chosen to list all, as I haven't seen any requirements for the kind of restriction you're suggesting. But I've made a note to myself to ask about this in today's meeting.

web/components/SessionCard/SessionCard.tsx
```diff
@@ -0,0 +1,40 @@
...
+  title,
+  startTime,
+  duration,
+  timezone,
+}: SessionCardProps) => (
+  <div className={styles.card}>
+    <strong className={styles.title}>{title}</strong>
+    {startTime && (
+      <div>
+        <time dateTime={startTime.toString()}>
```
oyvind-stenhaug, line 26: `toString` does not give a format that is valid here.

Originally it was implementation-dependent; now it'll give something like (depending on your timezone, and whether it's server-rendered or on the client)

```<time dateTime="Tue May 24 2022 19:30:00 GMT+0000 (Coordinated Universal Time)">Tuesday – May 24</time>```

HTML requires a particular format, which in this case should be a "[valid global date and time string](https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#valid-global-date-and-time-string)".

```suggestion
        <time dateTime={startTime.toISOString()}>
```

oyvind-stenhaug (changes requested):

web/pages/speakers/[slug].tsx
```diff
@@ -134,16 +145,41 @@ const SpeakersRoute = ({
...
+                sessions
+                  .filter(({ programContainingSession }) =>
+                    Boolean(programContainingSession)
+                  )
+                  .map(
+                    ({
+                      _id,
+                      title,
+                      programContainingSession: {
+                        startDateTime,
```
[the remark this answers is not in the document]
Comment (j0nas) [reply web/pages/speakers/[slug].tsx]: Agree with you, I didn't think to map to a better named field in the GROQ query. 

1c0b1f9. (Which also fixes the comment below, sorry, got carried away.)

web/util/queries.ts
```diff
@@ -49,6 +47,28 @@ const TICKET = `
...
+  title,
+  _createdAt,
+  _id,
+  _rev,
+  _type,
+  _updatedAt,
+  "sessions": *[_type == "session" && references(^._id)] {
+    _id,
+    title,
+    startTime,
```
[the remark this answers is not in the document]
j0nas: This should be fixed by d20e2e0, possibly with some relevant changes in subsequent commits.

web/pages/speakers/[slug].tsx
```diff
@@ -0,0 +1,213 @@
...
+interface SpeakersRouteProps {
+  data: {
+    speaker: Person & {
+      sessions?: {
+        title: string;
+        startTime: string;
+        duration: number;
+        timezone: string;
+        _id: string;
+      }[];
```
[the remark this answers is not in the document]
Comment (j0nas) [reply web/pages/speakers/[slug].tsx]: a3fb25c (with the enabling groundwork done in previous commits.)

web/hooks/useRandomShape.ts
j0nas: I don't know why I struggled with de-duplicating the shapes' classes before, but that's why they were left like that. A bit confused as to why, because converting _FakeItem.tsx_ to use the new hook was hassle-free: a4e9a63

oyvind-stenhaug:
I think this can be rubber-stamped as far as I'm concerned, but the PR description ought to be updated for future reference (and the PR marked as ready for review instead of draft, if no more changes are expected)

web/util/queries.ts
```diff
@@ -49,6 +47,38 @@ const TICKET = `
...
+  _createdAt,
+  _id,
+  _rev,
+  _type,
+  _updatedAt,
+  "sessions": *[_type == "session" && references(^._id) && !(_id in path("drafts.**"))] {
+    _id,
+    title,
+    duration,
+    "programContainingSession": *[_type == "program" && references(^._id)][0] {
```
oyvind-stenhaug, line 68: Nit: as discussed in Slack I'm not sure the default order is stable, so just in case this matches multiple programs it might be a good idea to use some explicit sort order (e.g. `_createdAt`)

j0nas: Good idea. edd5688.

j0nas:
@oyvind-stenhaug 

> [...] logic for calculating session start time [...] seems to use a "random" program instead of specifically looking for one that's connected to the "primary" venue, which was done on the session page.

As things are right now, I can't imagine a clean solution for this query with a "get the first venue tied to the main event" approach. Right now the query fetches the first program that references the queried session, if I switch it around then there's no guarantee that the first venue of the main event contains a program which holds the same reference, so I'd have to add a fallback case here. If there's no simpler solution to this, I think this might be a large amount of complexity for what currently seems as little gain? (Maybe I'm not seeing this right, correct me if so.) What do you think?

oyvind-stenhaug (approved):

PR #134, opened 2022-03-28
Session details page

j0nas:
# Session details page

## Intent

Page for individual Sessions, e.g. [Workshop Lunch][session]

## Description

We've scoped out the ["Session w/1 or 2 speakers"](https://www.figma.com/file/XjAvJKsLpxYJSc8QDWK81j/Sanity-SCC-Website-(Shared)?node-id=4476%3A16431) variant out of this PR; this PR only tackles the ["Session w/3+ speakers" sketches](https://www.figma.com/file/XjAvJKsLpxYJSc8QDWK81j/Sanity-SCC-Website-(Shared)?node-id=4476%3A16536).

## Testing this PR

1. Open a session, preferably one tied to a Speaker, e.g. [Workshop Lunch][session]
2. Verify that the page renders all fields correctly and appears in accordance with [the corresponding Figma sketches](https://www.figma.com/file/XjAvJKsLpxYJSc8QDWK81j/Sanity-SCC-Website-(Shared)?node-id=4476%3A16536) for all supported viewports

[session]: https://structured-content-2022-web-git-session-details-page.sanity.build/program/workshop-lunch


web/util/date.ts
```diff
@@ -2,24 +2,20 @@ import { formatInTimeZone } from 'date-fns-tz';
...
-  date: Date,
-  duration: number,
-  timezone: string
-) =>
-  `${formatTime(date, timezone)}–${formatTime(
-    addMinutes(date, duration),
-    timezone
-  )}`;
+export const formatTime = (date: Date, timezone: string, meridiem?: boolean) =>
+  formatInTimeZone(date, timezone, `HH:mm${meridiem ? ' aa' : ''}`, {
```
oyvind-stenhaug, line 6: This is mixing 24-hour format with AM/PM marker.

E.g. `formatTime(parseISO('2022-04-05T11:59Z'), 'Europe/Oslo', true)` shows "13:00 PM". Expected: "1:00 PM".

```suggestion
  formatInTimeZone(date, timezone, meridiem ? 'h:mm aa' : 'HH:mm', {
```

web/util/date.ts
```diff
@@ -30,6 +26,22 @@ export const formatTimeDuration = (start: Date, duration: number) => {
...
+  start: Date,
+  duration: number,
+  timezone: string
+) => {
+  const end = addMinutes(start, duration);
+  return `${formatTime(
+    start,
+    timezone,
+    differingMeridiem(start, end)
+  )}–${formatTime(end, timezone)}`;
```
oyvind-stenhaug, line 42: After the rewrite in 97d3dc91be877398e809fd2d48d4f42cf111065f, this doesn't show AM/PM at the end anymore.
```suggestion
  )}–${formatTime(end, timezone, true)}`;
```

oyvind-stenhaug, line 42: Minor thing, but per Carrie's comment there should be spaces around the en dash iff `differingMeridiem(start, end)` is true. It does perhaps look slightly confusing without, as if the "AM" "belongs" with the end time:

> 9:00 AM–1:00 PM

vs

> 9:00 AM – 1:00 PM

web/util/date.ts
```diff
@@ -30,6 +26,22 @@ export const formatTimeDuration = (start: Date, duration: number) => {
   return `PT${hours ? hours + 'H' : ''}${minutes ? minutes + 'M' : ''}`;
 };
 
+const differingMeridiem = (d1: Date, d2: Date) =>
+  d1.getHours() >= 12 !== d2.getHours() >= 12;
```
oyvind-stenhaug, line 30: This function doesn't take timezone into account. (This was an issue before the SessionDateTime rewrite too, but I unfortunately didn't catch it then.) How about comparing the outputs of `formatInTimeZone` with `aa` format token to see if they're the same or different, if there's no specialized function for it?

For instance, consider this function call:

```
formatTimeRange(parseISO('2022-04-05T09:00:00Z'), 240, 'America/Los_Angeles')
```

This currently yields " 02:00 AM–06:00" (at least on the server, which seems to use UTC, and on my client given that I'm in the Europe/Oslo timezone). After the fixes for my other comments it would show "2:00 AM – 6:00 AM". Expected: "2:00–6:00 AM".

Conversely, `formatTimeRange(parseISO('2022-04-05T18:00:00Z'), 180, 'America/Los_Angeles')` yields "11:00–14:00 PM" (or after accounting for the formatTime issue, "11:00–2:00 PM"), because in UTC/CEST the start/end times both have `getHours() >= 12`. Expected: "11:00 AM – 2:00 PM".

(I picked the examples carefully so that the bug would be reproducable both on server and on client in CEST, but in general the `getHours` use may cause a "Text content did not match" warning because of the different local timezones on client and server.)

oyvind-stenhaug (changes requested):
Any opinions on whether we should have a shared `SpeakerList` component for this page and the Speakers section?

There are quite a few similarities, although there are also some differences, so we'd need some sort of "variant" enum/bool prop:

Specifically, this page has
- smaller images on desktop
- different border-radius on the images
- Speaker role

If we're keeping them separate, I have some minor issues with the markup/styling. (And it might also be good to use similar mechanisms for both lists in the areas where they really are equal.) I can either comment or change it directly as we've done previously.

web/util/date.ts
j0nas: Good eye! :eyes: I'd missed this difference. 5cbdcb1

web/util/date.ts
j0nas: Wow. Amazing catch, and thank you very much for the fix suggestion as well. :heart: I don't see anything wrong with doing it this way. 9a514b4

I defaulted to `UTC` as `timeZone`. That doesn't matter, because the assumption is that both the compared dates originate from the same timezone, right?

j0nas:
@oyvind-stenhaug 

> If we're keeping them separate, I have some minor issues with the markup/styling. (And it might also be good to use similar mechanisms for both lists in the areas where they really are equal.) I can either comment or change it directly as we've done previously.

I saw somewhere that someone was using as a rule of thumb to have three instances of duplication before they refactored something out as a DRY class/component/whatever, and I liked this idea. Assuming such a rule of thumb (generally) makes sense, the lack of complete overlap at a mere two duplications would suggest that the resulting abstraction, if extracted, would perhaps be poor/insufficient/leaky/of low value/likely to diverge further/etc. I hope you see what I'm getting at. :slightly_smiling_face: So with that said, do you want to go ahead and fix the issues you've spotted directly?

web/util/date.ts
oyvind-stenhaug: No, I'm afraid it does matter, the point is that if `d1` is AM in UTC and `d2` is PM, that's not generally the case in the timezone we care about, and vice versa. It's a bit awkward to set up test content that demonstrates this, but maybe I should. The cases I mentioned above still apply.

I would think `differingMeridian` needs to take a timezone string as an argument. (Not a problem as such to make this parameter default to `UTC`, but `formatTimeRange` needs to pass its `timezone`.

web/util/date.ts
```diff
@@ -30,6 +26,22 @@ export const formatTimeDuration = (start: Date, duration: number) => {
...
+
+export const formatTimeRange = (
+  start: Date,
+  duration: number,
```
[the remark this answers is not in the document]
oyvind-stenhaug: Why would this function reverse it again? The only place I see it using `duration` is to compute an `end` variable.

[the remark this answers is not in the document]
j0nas: Sorry, poorly formulated. 

We receive a `start` Date and a `Duration` from Sanity. I believe it's more convenient to expose a util function to use those, rather than requiring callers to convert the `duration` into a Date.

oyvind-stenhaug (approved):

oyvind-stenhaug:
@j0nas Pushed my commits now.

The main user-visible change is that things used to kind of "jump around" as content was reflowed on hover. Also, there were double borders on pages where there were multiple speakers in the list. I used https://structured-content-2022-web-git-session-details-page.sanity.build/program/keynote as an example test for the latter.

Also adjusted the size of the arrow, and added the `:active` state/color from Figma as well.

We may want to adjust/fix the hover behavior of the speakers section to be consistent with this one, i.e. the borders being removed/obscured on a hovered item. This could be a small separate PR, though. Either way is fine with me.

oyvind-stenhaug:
(the last commits were OK-ed via Slack)

PR #135, opened 2022-03-31
Use data from Studio for primary navigation

j0nas:
# Use data from Studio for primary navigation

## Intent

Replaces the hardcoded items in _Nav.tsx_  with items fetched from Studio.
Solves [this](https://app.shortcut.com/sanity-io/story/16033/use-data-from-studio-primary-navigation-as-header-nav) Shortcut Story.

## Testing this PR

Step-by-step instructions which reviewers can use to verify that this PR
achieves its aim. Specify any prerequisites and assumptions. Be explicit.

1. Open the [Preview Deploy link](https://structured-content-2022-web-git-use-data-from-studio-pri-7f9396.sanity.build/) for this branch
2. Scroll down past the NavBlock to the point where the primary nav appears
3. Verify that the rendered items correspond with the data available in Studio
4. (optional) Modify the Primary navigation items and verify upon rebuilding the site that the updated links render in accordance with the Studio the changes you made

PR #137, opened 2022-04-01
Run Prettier on pre-commit, auto-sort imports

j0nas:
# Run Prettier on pre-commit, auto-sort imports

## Intent

As a response [this comment](https://github.com/sanity-io/structured-content-2022/pull/135#discussion_r839598600) (and because I consider having to remember to manually run `npx prettier --write ./` before every commit a bug): 
- set up [Husky](https://github.com/typicode/husky) and [lint-staged](https://github.com/okonet/lint-staged) to run [Prettier](https://github.com/prettier/prettier) on `pre-commit`
- set up [prettier-plugin-sort-imports](https://github.com/trivago/prettier-plugin-sort-imports) Prettier plugin to auto-sort imports

## Testing this PR

1. `yarn` to install new/updated dependencies, verify that the `prepare` npm script is successfully executed and that you get a message logged verifying that Husky git hooks were successfully installed
2. Perform a couple of test commits in various files, verify that your commits are formatted and that the imports are also sorted in a sensible order
    * _/web/pages/article/[slug].tsx_ is a good candidate for testing the sort ordering

### DIFF
--- a/web/.husky/pre-commit
+++ b/web/.husky/pre-commit
@@ -1,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+cd web && npx lint-staged
--- a/web/package.json
+++ b/web/package.json
@@ -5,7 +5,8 @@
     "dev": "next dev",
     "build": "next build",
     "start": "next start",
-    "lint": "next lint"
+    "lint": "next lint",
+    "prepare": "cd .. && husky install web/.husky"
   },
   "dependencies": {
     "@portabletext/react": "^1.0.2",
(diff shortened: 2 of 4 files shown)


oyvind-stenhaug (changes requested):
Starting to enforce a sort order for imports is fine by me, but I would strongly prefer if this PR is not merged without us also coordinating the reformatting of all the existing files. Getting all those extra diffs across various other PRs is a pain.

About the second part, i.e. reformatting on commit: Not ideal, but I suppose I can live with it. I tested a simple partial stage and it seemed to behave sensibly, which is nice. It seems to fail in more advanced scenarios, though, e.g. when staging only parts of a line. (It fails to restore the unstaged changes afterwards, and the commit is aborted.) It would have been nice to be able to opt out (e.g. if the hook install were separate from the regular `yarn` install), but I can understand if that's not possible with this tool.

Personally I've been using a small script (which a small personal hook reminded me to use; this hook is now overridden by husky) that ran `yarn lint` and `npx prettier -c .`. Is the `yarn lint` part included by the husky setup, or does it only run `prettier`?

PR #141, opened 2022-04-04
Sponsorship information: remove `available` field

j0nas:
# Sponsorship information: remove `available` field

## Intent

Fixes [this](https://app.shortcut.com/sanity-io/story/17275/remove-of-available-sponsorships-from-sponsorship-tiers) Shortcut Story.

## Description

Figma not fully updated at the time of submitting this PR, margins set based on [Slack clarifications](https://wja.slack.com/archives/C02E2ERKR7E/p1649066453854369?thread_ts=1649066287.323309&cid=C02E2ERKR7E).

## Testing this PR

1. Visit [/sponsorship-information](https://structured-content-2022-web-git-sponsorships-delete-avai-7b5db6.sanity.build/sponsorship-information)
2. Verify that no `available` field is rendered
3. Verify that all references to the field is deleted from the _/web_ directory


oyvind-stenhaug (approved):
Looks OK.

I see the Shortcut task also says "adjust heading and price so that they are aligned", which presumably refers to the fact that these table column headers have the default `vertical-align`.

We've already discussed that keeping both the names vertically aligned _and_ the prices vertically aligned seems like a huge hassle to do at best and not worth spending time on, but setting `vertical-align: top` seems better than ending up with `middle`. (Basically only relevant if the name is long enough that it wraps to multiple lines.)

Would be a very small fix, it's already set for `.rowHeader` so it would just be a matter of moving that declaration up to the `.columnHeader, .rowHeader` block.

j0nas:
@oyvind-stenhaug like 97d91a9?

oyvind-stenhaug:
@j0nas Yes, thank you :+1: 

The preview build does not seem to be updated with that change, I assume this is because of some Vercel breakage/downtime.

oyvind-stenhaug (approved):

PR #144, opened 2022-04-05
Delete revalidate endpoint

j0nas:
# Delete revalidate endpoint

## Intent

Delete no longer needed _/api/revalidate_ endpoint.

## Description

As the site has grown in complexity, there's now rarely a 1:1 correlation between one entity and one page, with most pages instead querying and rendering multiple entities. This breaks the simplistic assumption of our revalidation logic, which assumes that for any given mutated entity, there's a single page which needs to be revalidated. 

We could add complexity to the revalidation logic, but I think that would scale poorly and be needlessly complex compared to an alternative approach: the simplest way of ensuring the site is always up-to-date is to adopt sanity.io's approach of simply rebuilding the site when a change is picked up. A webhook has been set up in Studio which calls a Vercel webhook to rebuild and redeploy the site. This eliminates the need for this revalidate endpoint.

## Testing this PR

1. Verify that this change makes sense to you
2. Verify that all code related to this "revalidation" concept has been removed/disabled
3. (optional) Verify that the new Studio/Vercel webhooks which are to supersede this revalidation functionality works: make a small change in the [production Studio](https://admin.structuredcontent.live/) and verify that a new [Vercel build](https://vercel.com/sanity-io/structured-content-2022-web/deployments) is created, which eventually ships the change


j0nas:
@oyvind-stenhaug 

> Should the `@sanity/webhook` dependency be removed?

Good point, removed in b67b5bf.

oyvind-stenhaug (approved):

PR #147, opened 2022-04-06
Hotfix: program/[slug].tsx: fetch & pass Sanity nav items to <Nav />

j0nas:
# Hotfix: program/[slug].tsx: fetch & pass Sanity nav items to <Nav />

## Intent

Fixes build-breaking error in _program/[slug].tsx_.

## Testing this PR

1. Verify that CI passes
2. Verify that [/program](https://structured-content-2022-web-git-hotfix-program-page-pass-05cdfd.sanity.build/program) renders with main nav items from Sanity

### DIFF
--- a/web/pages/program/[slug].tsx
+++ b/web/pages/program/[slug].tsx
@@ -10,6 +10,7 @@ import TextBlock from '../../components/TextBlock';
 import { imageUrlFor } from '../../lib/sanity';
 import client from '../../lib/sanity.server';
 import { mainEventId } from '../../util/constants';
+import { PRIMARY_NAV } from "../../util/queries";
 import type { Slug } from '../../types/Slug';
 import type { Session } from '../../types/Session';
 import {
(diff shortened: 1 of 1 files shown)


web/pages/program/[slug].tsx
```diff
@@ -66,6 +69,7 @@ interface SessionRouteProps {
     home: {
       ticketsUrl: string;
     };
+    navItems: PrimaryNavItem[];
```
Comment (kmelve) [inline web/pages/program/[slug].tsx:72]: What happens if `navItems` is `null` or `undefined`?

kmelve (changes requested):
Looks great. Only one question about what happens if there's lacking data for `navItems` (Sanity returns `null` if there isn't any data)

oyvind-stenhaug (approved):
Approved since the breakage is blocking stuff, but the code should probably _also_ handle items being absent, along the lines of [Knut's PR](https://github.com/sanity-io/structured-content-2022/pull/150).

The query might not result in any items, in which case `null` is returned.

web/pages/program/[slug].tsx
Comment (j0nas) [reply web/pages/program/[slug].tsx]: Fixed in 619a555. :)

kmelve (approved):
LGTM!

PR #153, opened 2022-04-07
Sessions frontend, 1–2 speakers

oyvind-stenhaug:
# Sessions frontend, 1–2 speakers

## Intent

Implement the custom layout of speakers on the session details page in the case when there are only 1 or 2 speakers. Shortcut story [Session details page: front-end 1+2 speakers](https://app.shortcut.com/sanity-io/story/17072/sessions-details-page-front-end-1-2-speakers)

## Description

I had originally planned to adapt the `HighlightedSpeakerBlock` component for this use, but unfortunately there seems to be so many small differences in the Figma sketches that I ended up implementing a separate component instead (based on a copy/paste of the former).

As discussed in [Slack](https://wja.slack.com/archives/C02E2ERKR7E/p1649323933187139) the Figma sketches for desktop are significantly wider than the narrowest viewport that qualifies, so this page uses a custom breakpoint to try to ensure that there's actually enough room for two images side by side (without having to severely shrink them, which also affects the text blocks within that show the speakers' name/title/company). (This also contributes to the divergence of HighlightedSpeakerBlock and SessionSpeakers.)

## Testing this PR

1. Open https://structured-content-2022-web-git-sessions-frontend-1-2-speakers.sanity.build/program/cross-functional-collaboration-for-structured-content  (2 speakers), verify that the speaker presentation looks OK and in accordance with the Figma sketch (linked in Shortcut) 
2. Open https://structured-content-2022-web-git-sessions-frontend-1-2-speakers.sanity.build/program/workshop-lunch (1 speaker), verify that the speaker presentation looks OK

### DIFF
--- a/web/pages/program/[slug].tsx
+++ b/web/pages/program/[slug].tsx
@@ -196,8 +196,8 @@ const SessionRoute = ({
                 (hasHighlightedSpeakers ? (
                   <div className={programStyles.highlightedSpeakers}>
                     <SessionSpeakers
-                      photo={speakers[0]?.person?.photo}
-                      photo2={speakers[1]?.person?.photo}
+                      speaker1={speakers[0]?.person}
+                      speaker2={speakers[1]?.person}
                     />
                   </div>
                 ) : (
--- a/web/pages/program/[slug].tsx
+++ b/web/pages/program/[slug].tsx
@@ -1,10 +1,12 @@
 import type { GetStaticPaths, GetStaticProps } from 'next';
 import { groq } from 'next-sanity';
+import clsx from 'clsx';
 import urlJoin from 'proper-url-join';
 import Footer from '../../components/Footer';
 import GridWrapper from '../../components/GridWrapper';
 import MetaTags from '../../components/MetaTags';
 import Nav from '../../components/Nav';
+import SessionSpeakers from '../../components/SessionSpeakers';
 import Tag from '../../components/Tag';
 import TextBlock from '../../components/TextBlock';
 import { imageUrlFor } from '../../lib/sanity';
@@ -144,6 +146,8 @@ const SessionRoute = ({
     currentSessionInProgram?._id,
     sessions
   );
+  const hasHighlightedSpeakers =
+    speakers?.length === 1 || speakers?.length === 2;
   return (
     <>
       <MetaTags title={title} description="" currentPath={`/session/${slug}`} />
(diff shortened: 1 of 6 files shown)


web/pages/program/[slug].tsx
```diff
@@ -147,45 +192,17 @@ const SessionRoute = ({
...
-                      </li>
-                    )
-                  )}
-                </ul>
-              )}
+              {speakers &&
+                (hasHighlightedSpeakers ? (
+                  <div className={programStyles.highlightedSpeakers}>
+                    <SessionSpeakers
+                      speaker1={speakers[0]?.person}
```
Comment (j0nas) [inline web/pages/program/[slug].tsx:199]:
```suggestion
                      speaker1={speakers[0].person}
```

(Super-nit, but since `hasHighlightedSpeakers` is true in this branch, `speakers[0]` can't be false. Perhaps then maybe we should remove the optional flag for `speaker1` in `<SessionSpeakers />` because it can't ever render with a falsy `speaker1`.)

web/styles/media.css
```diff
@@ -1,3 +1,4 @@
 @custom-media --small-up (min-width: 390px);
 @custom-media --medium-up (min-width: 768px);
 @custom-media --large-up (min-width: 1025px);
+@custom-media --session-large-up (min-width: 1310px);
```
j0nas, line 4: Why is there a custom breakpoint defined specifically for the Sessions page? Can't the Figma sketches be adjusted to be consistent with the conventional breakpoints?

j0nas (approved):
Great job, this looks very good to me. I left a couple nits, but all of it is very optional. :slightly_smiling_face:

web/pages/program/[slug].tsx
Comment (oyvind-stenhaug) [reply web/pages/program/[slug].tsx:201]: Technically speaking, `person` is apparently not a mandatory field, so it's possible to publish a session that results in us receiving something like `"speakers":[{"person":null,"role":"moderator"}]`.

Good point about `speakers[0]` not being falsy, though. I went through one or two iterations of the `hasHighlightedSpeakers` check, and this wasn't originally the case.

web/styles/media.css
oyvind-stenhaug: This is explained in the PR description ("Description" section, 2nd paragraph).

Originally, I intentionally set the breakpoints so that there would only ever be _more_ space available than in Figma. But this meant the "tablet" design applied to a lot of configurations, so it was adjusted based on Sindre's input. (Also, there's no sketch for <390px anyway.)

I don't think it's so easy to adjust the desktop sketch to work down to 1025px without basically throwing out/redoing most of it, and I suppose it's already been approved, so that's why we ended up with this. Maybe sometime after Easter we should discuss a somewhat less waterfall-y approach to things (also it would be nice if the desktop sketch was made at 1025px width, and similarly for the other variants, but this is presumably a big job to re-do now…)

oyvind-stenhaug:
I will merge this since all of the changes done after approval are from GitHub suggestions accepted verbatim.

PR #160, opened 2022-04-20
Additional metadata for Speakers, Sessions pages

j0nas:
# Additional metadata for Speakers, Sessions pages

## Intent

Solves [this](https://app.shortcut.com/sanity-io/story/17753/generate-og-images-for-speaker-pages-based-on-speaker-photo-where-available-for-previews-on-facebook-twitter-etc) Shortcut Story.

## Description

- Pass Speaker's `photo` to `<MetaTags />` in _/speakers/[slug].tsx_
- If associated with a Program, also pass Session's start date, time, duration for SF timezone in _/program/[slug].tsx_

## Testing this PR

1. Go to https://metatags.io/ and enter https://structured-content-2022-web-git-metadata-for-speakers-sessions.sanity.build/speakers/simen-svale-skogsrud in the input field at the top of the page. (The page is loaded on the field's `onblur` event, it seems; click somewhere else after inputting the field to load the contents.) Verify that the photo of Simen is the one that's loaded, as opposed to the default image that's currently shown for https://structuredcontent.live/speakers/simen-svale-skogsrud.
2. Still at metatags.io, enter a Session that's associated with the main San Francisco venue, e.g. https://structured-content-2022-web-git-metadata-for-speakers-sessions.sanity.build/program/workshop-2022. Verify that _Tuesday, May 24 9:00 AM – 12:30 PM PDT_ is displayed in the meta description field.
3. Enter a Session that's not associated with the SF program, e.g. https://structured-content-2022-web-git-metadata-for-speakers-sessions.sanity.build/program/sessions-2. Verify that you can still access the page for the session without encountering any errors or warnings, and that metatags.io render's the respective page's description field as empty.

PR #162, opened 2022-04-22
Session details: show dates & times for all Venues referencing Session

j0nas:
# Session details: show dates & times for all Venues referencing Session

## Intent

List date and time for current Session for each Program/Venue combo which it appears in.
image

Solves [this](https://app.shortcut.com/sanity-io/story/17644/multiple-times-on-session-detail) Shortcut Story.

## Description

- Fetch all Venues from the main event
- Fetch Programs referencing to said Venues, in order to have Programs match the main event's Venue sort order
- For each Program referencing the current Session, render date and time info

Note that this PR [also fixes what I believe is a bug](https://github.com/sanity-io/structured-content-2022/commit/e3c25a0d25eb1ae8d11dd590b2604ba6c83fda0c) in _date.ts_#`formatTime`. This affects how the [Session times in the accordion under the "Programs: section heading"](https://structured-content-2022-web-git-session-details-multiple-times.sanity.build/program) render. This fix looks correct to me (and [the current production version](https://structuredcontent.live/program) does not, but please verify this.

(Note also that the code assumes that each Venue is only tied to a single Program, despite the Sanity model expressing the Venue->Program relationship as a one-to-many. Knut has previously mentioned somewhere on Slack that this will in in reality always be a one-to-one relationship.)

## Testing this PR

1. Open a Session that's referenced by at least one Program (which in turn is tied to a Venue referenced by the main event), e.g. [Cross-Functional Collaboration for Structured Content](https://structured-content-2022-web-git-session-details-multiple-times.sanity.build/program/cross-functional-collaboration-for-structured-content)
2. Verify that the dates & times for the event correlate with the data that's in Sanity for the event; make sure that both the displayed values and the `datetime` values of the respective tags are correct.
3. Open a Session that doesn't meet the criteria of 1) -- e.g. [Session not referenced by any Programs](https://structured-content-2022-web-git-session-details-multiple-times.sanity.build/program/not-referenced). Verify that it omits rendering any timing information without encountering any errors.

### DIFF
--- a/web/pages/program/[slug].tsx
+++ b/web/pages/program/[slug].tsx
@@ -23,8 +23,6 @@ import type { Session } from '../../types/Session';
 import { sessionTimingDetailsForMatchingPrograms } from '../../util/session';
 import styles from '../app.module.css';
 import programStyles from './program.module.css';
-import { formatTimeDuration } from '../../util/date';
-import { parseISO } from 'date-fns';
 
 const QUERY = groq`
   {
--- a/web/pages/program/[slug].tsx
+++ b/web/pages/program/[slug].tsx
@@ -173,9 +173,10 @@ const SessionRoute = ({
                   ({ time, label, timezone, date }) => (
                     <p key={time}>
                       <strong>{label}</strong>
-                      {/* should these be <time> tags? */}
-                      <span style={{ display: 'block' }}>{date}</span>
-                      <span style={{ display: 'block' }}>
+                      <span className={programStyles.sessionVenueDateTime}>
+                        {date}
+                      </span>
+                      <span className={programStyles.sessionVenueDateTime}>
                         {time} {timezone}
                       </span>
                     </p>
--- a/web/pages/program/[slug].tsx
+++ b/web/pages/program/[slug].tsx
@@ -23,6 +23,8 @@ import type { Session } from '../../types/Session';
 import { sessionTimingDetailsForMatchingPrograms } from '../../util/session';
 import styles from '../app.module.css';
 import programStyles from './program.module.css';
+import { formatTimeDuration } from '../../util/date';
+import { parseISO } from 'date-fns';
 
 const QUERY = groq`
   {
@@ -170,15 +172,21 @@ const SessionRoute = ({
                 <h1 className={programStyles.sessionTitle}>{title}</h1>
 
                 {matchingSessionsInPrograms.map(
-                  ({ time, label, timezone, date }) => (
-                    <p key={time}>
+                  ({ label, time, timezone, rawDate, date, duration }) => (
+                    <p key={label}>
                       <strong>{label}</strong>
-                      <span className={programStyles.sessionVenueDateTime}>
+                      <time
+                        dateTime={rawDate}
+                        className={programStyles.sessionVenueDateTime}
+                      >
                         {date}
-                      </span>
-                      <span className={programStyles.sessionVenueDateTime}>
+                      </time>
+                      <time
+                        dateTime={duration}
+                        className={programStyles.sessionVenueDateTime}
+                      >
                         {time} {timezone}
-                      </span>
+                      </time>
                     </p>
                   )
                 )}
--- a/web/pages/program/[slug].tsx
+++ b/web/pages/program/[slug].tsx
@@ -51,6 +51,7 @@ const QUERY = groq`
     "programs": *[_id == "${mainEventId}"].venues[]-> {
       "program": *[_type == "program" && references(^._id)] { ${PROGRAM} }
     }["program"][],
+    "mainVenueName": (*[_id == "${mainEventId}"].venues[0]->.name)[0],
   }`;
 
 interface SessionRouteProps {
@@ -76,6 +77,7 @@ interface SessionRouteProps {
       };
     };
     programs: Program[];
+    mainVenueName: string;
   };
   slug: string;
 }
@@ -128,6 +130,7 @@ const SessionRoute = ({
     navItems,
     footer,
     programs,
+    mainVenueName,
   },
   slug,
 }: SessionRouteProps) => {
@@ -139,7 +142,7 @@ const SessionRoute = ({
     _id
   ).map(({ label, ...rest }) => ({
     // Ad-hoc override for the SF venue, for this specific view only
-    label: label === 'San Francisco' ? 'San Francisco & Virtual' : label,
+    label: label === mainVenueName ? 'San Francisco & Virtual' : label,
     ...rest,
   }));
   return (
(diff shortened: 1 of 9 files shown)


web/util/date.ts
```diff
@@ -3,7 +3,7 @@ import { addMinutes, intervalToDuration } from 'date-fns';
 import enUS from 'date-fns/locale/en-US';
 
 export const formatTime = (date: Date, timezone: string, meridiem?: boolean) =>
-  formatInTimeZone(date, timezone, meridiem ? 'h:mm aa' : 'HH:mm', {
+  formatInTimeZone(date, timezone, `h:mm${meridiem ? ' aa' : ''}`, {
```
oyvind-stenhaug, line 6: I think the confusion that arose regarding `meridiem` suggests that the parameter name is too vague. I think it would be good to rename it and/or to add function documentation[1].

Second, given that it is now clear that the parameter isn't a toggle between 12-hour and 24-hour anymore (a change that makes sense), I really think the default should be to _show_ the AM/PM marker, so that the caller must explicitly opt in to the ambiguous rendering.

Combing the two comments above, I would suggest replacing `meridiem?` with something like `hideMeridiemIndicator?`, inverting the logic.

[1] We have few of those now, but for programs of a certain size/complexity I really prefer to be able to read some JSdoc or equivalent, rather than having to read implementation details several levels deep in order to try to piece together some implied contract.

web/util/session.ts
```diff
@@ -25,3 +32,47 @@ export const sessionStart = (
...
   return addMinutes(start, sessionStartOffset);
 };
+
+const getDuration = ({ session, duration, durationOverride }: ProgramSession) =>
+  durationOverride ?? duration ?? session?.duration ?? 0;
+
+/* Given all Programs, finds the Programs that contain the Session (matching on _id).
+ * For each match, returns:
+ * - the Program's associated Venue name (using the first entry in the Program's Venues array)
+ * - the Session's date (based on the Session's start time)
```
oyvind-stenhaug, line 42: As commented in a "footnote" above, I appreciate the function comment.

This doesn't mention `rawDate`, though.

oyvind-stenhaug (changes requested):
(It's odd that it's showing "GMT+2" instead of "CEST" as I would expect, but this seems to be due to the `en-US` locale, unfortunately. And it seems a bit hacky to set `en-GB` in the `getNonLocationTimezone` function just to work around that.)

Regarding `formatTime`: I'm pretty sure the Program page used to be 24-hour format intentionally, but the Figma sketches probably changed at some point due to feedback from Sanity. And it makes perfect sense to consistently use 12-hour format everywhere, so the change to fix `formatTimeRange` looks good to me. Left a couple comments inline, though.

web/pages/program/[slug].tsx
```diff
@@ -136,25 +123,25 @@ const SpeakerList = ({ speakers }: SpeakerListProps) => (
...
-  );
   const hasHighlightedSpeakers =
     speakers?.length === 1 || speakers?.length === 2;
+
+  const matchingSessionsInPrograms = sessionTimingDetailsForMatchingPrograms(
+    programs,
+    _id
+  ).map(({ label, ...rest }) => ({
+    // Ad-hoc override for the SF venue, for this specific view only
+    label: label === 'San Francisco' ? 'San Francisco & Virtual' : label,
```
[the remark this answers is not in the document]
Comment (j0nas) [reply web/pages/program/[slug].tsx]: Like you mentioned on Slack, the waterfall-y nature of this project sometimes leads to these kinds of solutions. I don't really want to modify the schema, so in my quest to make Figma-compliant implementations, these types of hacks sometimes take place.

I followed your "main venue" suggestion to the best of my understanding in d6fad47.

web/pages/program/[slug].tsx
```diff
@@ -182,20 +169,24 @@ const SessionRoute = ({
...
-                    <div>
-                      {formatTimeRange(
-                        start,
-                        currentSessionInProgram.duration,
-                        mainVenueTimezone
-                      )}{' '}
-                      {getNonLocationTimezone(start, mainVenueTimezone, true)}
-                    </div>
-                  </>
+                {matchingSessionsInPrograms.map(
```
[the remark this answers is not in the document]
Comment (j0nas) [reply web/pages/program/[slug].tsx]: ee4cc82.

web/pages/program/[slug].tsx
```diff
@@ -182,20 +169,24 @@ const SessionRoute = ({
...
+                  ({ label, time, timezone, rawDate, date, duration }) => (
+                    <p key={label}>
+                      <strong>{label}</strong>
+                      <time
+                        dateTime={rawDate}
+                        className={programStyles.sessionVenueDateTime}
```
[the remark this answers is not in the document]
Comment (j0nas) [reply web/pages/program/[slug].tsx]: 66e98a2.

web/util/date.ts
j0nas: 2e239a8.

web/util/session.ts
j0nas: 8fd3aca.

web/util/session.ts
```diff
@@ -25,3 +32,48 @@ export const sessionStart = (
...
   return addMinutes(start, sessionStartOffset);
 };
+
+const getDuration = ({ session, duration, durationOverride }: ProgramSession) =>
+  durationOverride ?? duration ?? session?.duration ?? 0;
+
+/* Given all Programs, finds the Programs that contain the Session (matching on sessionId === _id).
+ * For each match, returns:
+ * - label: Program's associated Venue name (using the first entry in the Program's Venues array)
+ * - rawDate: the Session's date (based on the Session's start time)
```
oyvind-stenhaug, line 42: Actually I'm really confused to what `rawDate` is supposed to be now. It's not really a "date", it's a moment in time as a string on a particular ISO format). It seems to be the start time of the _program_? Is this just a bug?

Example: cross-functional-collaboration-for-structured-content

```
<time datetime="2022-05-24T15:00:00.000Z">Tuesday, May 24</time>
<time datetime="PT3H30M">12:30–4:00 PM PDT</time>
```

So the machine-readable info seems to be just the moment in time when the _program_ starts and the duration of the session. What would make more sense to me—as long as it's giving one moment in time and one duration—is to have the moment in time when the _session_ starts (Tuesday, May 24, 12:30 PM PDT), and its duration.

```
<time datetime="2022-05-24T19:30:00.000Z">Tuesday, May 24</time>
<time datetime="PT3H30M">12:30–4:00 PM PDT</time>
```

oyvind-stenhaug (changes requested):

web/util/session.ts
j0nas: This.. seems to have been a somewhat significant bug. Fixed in 9e7120c. Once the `jest` setup from #163 lands, I'd like to use that to write some tests for this. Good catch.

j0nas:
Same as with #160: merging per Wilhelm's request.

PR #166, opened 2022-04-28
Link to Sessions in Speaker page

j0nas:
# Link to Sessions in Speaker page

## Intent

Wrap Speaker's listed Sessions in links.

## Description

In order to do this a bit more _properly_, (e.g. using `getEntityPath`), a bit of a refactor was required. As part of this, I also rewrote some types and queries from ad-hoc ones into reusing existing types/queries.

(E.g. the `${SESSION}` query rewrite leads to more bytes queried, but this has no impact on the end user, as these pages are pre-generated, so the cleaner/DRYer code is a win, IMO.)

## Testing this PR

1. Go to [/speakers](https://structured-content-2022-web-git-speaker-page-link-to-session.sanity.build/speakers)
2. For each accessible Speaker from the list, verify that the Speaker's Sessions are listed the same way, (i.e. all Session entries match that of those currently listed for `main`, so that the refactor in this PR doesn't break existing logic) and that they all correctly link to respective Sessions


web/util/queries.ts
```diff
@@ -96,18 +96,15 @@ const SPEAKER_WITH_SESSIONS = `
   _type,
   _updatedAt,
   "sessions": *[_type == "session" && references(^._id) && !(_id in path("drafts.**"))] {
-    _id,
-    title,
-    duration,
+    "session": { ${SESSION} },
```
oyvind-stenhaug, line 99: This should only query for the fields we actually need. If it helps, we could have a separate `SPEAKER_SESSION` const or something like that.

> the ${SESSION} query rewrite leads to more bytes queried, but this has no impact on the end user, as these pages are pre-generated

No, as far as I can tell, this is not the case. All this ends up being downloaded and processed by the client for "hydration" purposes. If I compare the size of the HTML page for Andy Fitzgerald's speaker page on `main` vs the `speaker-page-link-to-session` branch, it's well over 2 times as large now. The majority of it is a huge JSON blob at the end.

oyvind-stenhaug (changes requested):

web/components/SessionCard/SessionCard.module.css
```diff
@@ -7,6 +7,18 @@
   letter-spacing: var(--letter-spacing-ui-xs);
 }
 
+.link {
+  text-decoration: none;
+}
```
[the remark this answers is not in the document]
j0nas: b96af0a.

web/components/SessionCard/SessionCard.tsx
```diff
@@ -1,40 +1,53 @@
...
   formatTimeDuration,
   formatTimeRange,
   getNonLocationTimezone,
 } from '../../util/date';
+import { getEntityPath } from '../../util/entityPaths';
 import styles from './SessionCard.module.css';
 
-interface SessionCardProps {
-  title: string;
+interface SessionCardProps extends Session {
```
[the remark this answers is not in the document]
j0nas: 82ebc97.

web/components/SessionCard/SessionCard.tsx
```diff
@@ -1,40 +1,53 @@
...
   sessionStart?: Date;
-  duration?: number;
   timezone?: string;
 }
 
 export const SessionCard = ({
+  _type,
```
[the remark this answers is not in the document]
j0nas: I pass this `_type` down to `getEntityPath` rather than hardcoding it to stay consistent with the rest of the codebase.

web/util/queries.ts
j0nas: 2fe0413

web/pages/speakers/[slug].tsx
```diff
@@ -39,16 +41,10 @@ const QUERY = groq`
...
-  duration: number;
+  session: Pick<Session, '_id' | '_type' | 'title' | 'duration' | 'slug'>;
   programContainingSession: {
     programStart: string;
-    sessions: {
-      _type: string;
-      duration: number;
-      session?: Pick<Session, '_id' | 'duration'>;
-    }[];
+    sessions: ProgramSession[];
```
Comment (oyvind-stenhaug) [inline web/pages/speakers/[slug].tsx:47]: Nit: This type implies that any non-null `session` within will have a `_type`, which is not the case. The query only fetches an `_id`.

web/util/queries.ts
```diff
@@ -96,18 +96,15 @@ const SPEAKER_WITH_SESSIONS = `
...
       sessions[] {
+        _key,
         _type,
         duration,
-        session-> {
-          _id,
-          duration,
-        }
+        durationOverride,
+        session->{ _id },
```
oyvind-stenhaug, line 107: I find the session timing code quite difficult to follow with all the nested session objects (due in no small part to the complexity of the program/session setup in the underlying schema, I'm sure), but I believe the removal of `duration` here (in commit https://github.com/sanity-io/structured-content-2022/commit/2fe04133f72fbf0282422eaf57d21511b6c8e385) has caused the listed session times to break.

It looks like they are now ignoring the durations of all preceding sessions, so that only "padding" entries are taken into account.

E.g. on http://localhost:3000/speakers/carrie-hane:
- "The Future of Digital Experience […]": used to show "1:15–2:05 PM PDT", now shows "8:30–9:20 AM PDT" (8:30 is the start time of the program, but the session has 4 sessions before it, so it should start 4h 45min later)
- "Improv Consulting […]" used to show "3:15–4:00 PM PDT" but now shows "9:05–9:50 PM PDT" (which seems offset by the preceding padding entries, but not by the preceding sessions)

oyvind-stenhaug (changes requested):

web/util/queries.ts
oyvind-stenhaug: https://structured-content-2022-ggd84iq0r.sanity.build/speakers/carrie-hane (before)
vs https://structured-content-2022-lgit97v4t.sanity.build/speakers/carrie-hane (after)

j0nas: Yeah, the session-handling code is a bit out of control at this point. One factor contributing to this is the schema, as you mention, wrt. both forcing clients to calculate start/end times and the session-in-session structure where a program has a "session" object which either is a "padding"-type session, or wraps an actual Session. Additionally, The lack of any kind of ORM/codegen/strongly typed client which we could use rather than defining our own types that end up being "out of sync" with the schema is also making things more difficult.

In 799a9cc, I attempt to simplify things a bit. This simplification could probably also be ported over to the Program page, but I didn't want to increase the scope of this PR.

oyvind-stenhaug: Thanks, that looks better :+1: And the session start time is fixed.

PR #167, opened 2022-04-29
On-the-fly fallback og:images

j0nas:
# On-the-fly fallback og:images

## Intent

- Exposes an API endpoint for generating fallback `og:image`s based on the slug on which the endpoint is accessed
- Sets fallback `og:image` image for every route in application

Solves [this](https://app.shortcut.com/sanity-io/story/18784/site-page-og-images) Shortcut Story.

## Description

This PR uses a base image stolen from @oyvind-stenhaug [test branch](https://github.com/sanity-io/structured-content-2022/compare/oyvinds/og-image-test) and overlays the specified text on top of it using [jimp](https://github.com/oliver-moran/jimp/tree/master/packages/jimp) (which does not require a headless browser :tada:). I also included the [word-wrap](https://github.com/jonschlinkert/word-wrap) package to wrap long strings.

I used a web service to convert the "national-2-narrow-bold" woff2 font to a .ttf, then used [TTF to FNT](https://ttf2fnt.com/) to convert it again to a _.fnt + .bmp_ as required by `jimp`. `jimp` exposes functionality for calculating a string's resulting dimensions, but I've opted for hardcoded values, as the fnt + .bmp_ files are fixed to 80px. I also compressed both .bmp files introduced in this PR.)

I deleted the no-longer-needed [opengraph-image](https://user-images.githubusercontent.com/6001131/165936194-a032331d-9084-4e72-88ff-f2529844376a.png), but am including a link to it here so that it can be uploaded as the image-to-be-used for the root/homepage route in Sanity.

Two things to note with this PR:
- I haven't tested the performance of this, but in theory this PR could expose a DoS vulnerability due to images being generated on every request without any form of cache being involved. Since we're not using a headless browser to generate the images, the solution in this PR is far more performant than alternatives, but it is possible that frequent calls to this endpoint in theory could prove computationally expensive. Again, I haven't done any testing, but worth keeping this point in mind in case we want to improve on this in the future
- I haven't added any restrictions wrt. which input strings the API accepts in this first iteration; any alphanumeric (plus some special characters) input will yield an image with the given text. If deemed necessary, we could query Sanity for valid fields which we could restrict the output with, but this would add more overhead and complexity which I hope to scope out of the first iteration if at all possible

## Testing this PR

Use a SEO inspection tool, e.g. [opengraph.xyz](https://structured-content-2022-web-git-generate-og-image.sanity.build/program/), to visit various routes in the [Preview Deploy](https://structured-content-2022-web-git-generate-og-image.sanity.build/). Verify that routes which have images uploaded in their respective _SEO_ sections in Sanity, render those images in the `og:image` meta tags, and that Speaker routes render the images of the respective Speakers, where available, and finally verify that the `og:image`s introduced by this PR are rendered where no other relevant images apply.


web/pages/api/og-image/[slug].ts
```diff
@@ -0,0 +1,42 @@
...
+    const image = await Jimp.read(imagePath);
+
+    const font = await Jimp.loadFont(
+      resolve('public', 'static', 'fonts', 'national-2-narrow-bold.fnt')
+    );
+    wrap(req.query.slug as string, { width: WORD_WRAP_LIMIT })
+      .split('\n')
+      .forEach((line, i) =>
+        image.print(font, PADDING_X, PADDING_Y + i * FONT_SIZE, line)
+      );
```
Comment (oyvind-stenhaug) [inline web/pages/api/og-image/[slug].ts:30]: I don't understand why this is wrapped based on number of characters. This means the lines can get extra "jagged" if we have bad "luck" with the string contents, since the font is not fixed width. Also it doesn't prevent the text from bleeding over the shapes on the right, which the Figma sketches do try to avoid.

Is it because of the line height? And if so, could it be an option to modify the .fnt file's `lineHeight` instead so that Jimp's own text wrapping can be used instead of the `wrap` + `split` + `forEach` combination?

E.g. `image.print(font, PADDING_X, PADDING_Y, req.query.slug as string, 752);` (where 752 can be tweaked if necessary, and of course be defined in a constant)

oyvind-stenhaug (changes requested):

web/pages/speakers/[slug].tsx
```diff
@@ -18,6 +18,7 @@ import client from '../../lib/sanity.server';
 import { mainEventId, newsletterSharedSectionId } from '../../util/constants';
 import twitterLogo from '../../images/twitter_logo_black.svg';
 import linkedinLogo from '../../images/linkedin_logo_black.svg';
+import { getOgImagePath } from "../../util/entityPaths";
```
Comment (oyvind-stenhaug) [inline web/pages/speakers/[slug].tsx:21]:
```suggestion
import { getOgImagePath } from '../../util/entityPaths';
```

web/pages/articles/[slug].tsx
```diff
@@ -13,6 +13,7 @@ import TextBlock from '../../components/TextBlock';
 import client from '../../lib/sanity.server';
 import { formatDate } from '../../util/date';
 import { mainEventId } from '../../util/constants';
+import { getOgImagePath } from "../../util/entityPaths";
```
Comment (oyvind-stenhaug) [inline web/pages/articles/[slug].tsx:16]:
```suggestion
import { getOgImagePath } from '../../util/entityPaths';
```

web/pages/speakers/[slug].tsx
Comment (j0nas) [reply web/pages/speakers/[slug].tsx]: 30a261d.

web/pages/api/og-image/[slug].ts
```diff
@@ -0,0 +1,42 @@
...
+  req: NextApiRequest,
+  res: NextApiResponse
+) {
+  try {
+    // Resolves to root of /web directory, __dirname is buggy in Next.js
+    const imagePath = resolve('images', 'og_image_background.png');
+    const image = await Jimp.read(imagePath);
+
+    const font = await Jimp.loadFont(
+      resolve('public', 'static', 'fonts', 'national-2-narrow-bold.fnt')
```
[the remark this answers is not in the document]
Comment (j0nas) [reply web/pages/api/og-image/[slug].ts]: 9e81d2c.

web/pages/articles/[slug].tsx
Comment (j0nas) [reply web/pages/articles/[slug].tsx]: 30a261d.

web/pages/[[...slug]].tsx
```diff
@@ -203,6 +204,7 @@ const Route = ({ data: initialData, slug, preview }: RouteProps) => {
     <>
       <MetaTags
         {...{ title, description: seoDescription, image, currentPath, noIndex }}
+        fallbackImage={{ url: getOgImagePath(title), alt: title }}
```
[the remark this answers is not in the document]
Comment (j0nas) [reply web/pages/[[...slug]].tsx]: 31724d2.

web/pages/api/og-image/[slug].ts
Comment (j0nas) [reply web/pages/api/og-image/[slug].ts]: I didn't study `jimp`'s docs closely enough to notice that it had a built-in word-wrapping option. Good catch! 07edc7d.

Comment (oyvind-stenhaug) [reply web/pages/api/og-image/[slug].ts]: OK.

As mentioned, it seems to be possible to adjust `lineHeight=96` to e.g. `lineHeight=77` in the .fnt file to decrease the line height when there are many lines, e.g. https://structured-content-2022-web-git-generate-og-image.sanity.build/api/og-image/How%20Publishing%20Organizations%20Harness%20Structured%20Content%20to%20Power%20Editorial%20Experiences.

But this is just based on experiment, I haven't read up on the specs for .fnt if there is such a thing.

oyvind-stenhaug (approved):
LGTM

web/pages/api/og-image/[slug].ts
Comment (j0nas) [reply web/pages/api/og-image/[slug].ts]: Good suggestion, that looks much better. 2584d37

PR #170, opened 2022-05-08
Show Venue address in Program

j0nas:
# Show Venue address in Program

## Intent

When viewing a Program for a specific Venue, show the selected Venue's address under the VenueNav.

## Details

Carrie has [approved](https://wja.slack.com/archives/C02UKELNTE0/p1652104017244119?thread_ts=1651852344.287539&cid=C02UKELNTE0) these changes.

## Testing this PR

1. Open [/program](https://structured-content-2022-web-git-program-add-venue-address.sanity.build/program)
2. Navigate between the available venues using the venue nav
3. Verify that the address renders correctly for each respective nav, if present in the dataset

### DIFF
--- a/web/components/TextBlock/Programs/Programs.module.css
+++ b/web/components/TextBlock/Programs/Programs.module.css
@@ -3,7 +3,6 @@
 }
 
 .venueAddress {
-  text-align: center;
   margin: 0;
 }
 
@@ -85,6 +84,10 @@
     display: flex;
     justify-content: center;
     transform: translateY(-50%);
+  }
+
+  .venueAddress {
+    text-align: center;
   }
 
   .container {
--- a/web/components/TextBlock/Programs/Programs.module.css
+++ b/web/components/TextBlock/Programs/Programs.module.css
@@ -1,9 +1,14 @@
 .venueNavContainer {
-  margin: 32px 0;
+  margin: 32px 0 48px;
+}
+
+.venueAddress {
+  text-align: center;
+  margin: 0;
 }
 
 .container {
-  margin: 64px 0 80px;
+  margin: 48px 0 80px;
 }
 
 .heading {
@@ -76,14 +81,14 @@
 
 @media (--medium-up) {
   .venueNavContainer {
-    margin: 0;
+    margin: 0 0 calc(64px - (77px / 2)); /* 77px is the height of VenueNav */
     display: flex;
     justify-content: center;
     transform: translateY(-50%);
   }
 
   .container {
-    margin: 60px 0 96px;
+    margin: 64px 0 96px;
   }
 
   .dayHeader {
--- a/web/components/TextBlock/Programs/Programs.tsx
+++ b/web/components/TextBlock/Programs/Programs.tsx
@@ -148,12 +148,12 @@ export const Programs = ({
 
         {typeof address === 'object' && (
           <address className={styles.venueAddress}>
-            <div>{address.name}</div>
-            <div>{address.street}</div>
+            <div>{address?.name}</div>
+            <div>{address?.street}</div>
             <div>
               {[
-                address.city,
-                [address.state, address.postalCode].filter(Boolean).join(' '),
+                address?.city,
+                [address?.state, address?.postalCode].filter(Boolean).join(' '),
               ]
                 .filter(Boolean)
                 .join(', ')}
--- a/web/components/TextBlock/Programs/Programs.tsx
+++ b/web/components/TextBlock/Programs/Programs.tsx
@@ -1,6 +1,7 @@
 import { parseISO } from 'date-fns';
 import Link from 'next/link';
 import { useRouter } from 'next/router';
+import { Fragment } from 'react';
 import type { PortableTextComponentProps } from '@portabletext/react';
 import type { EntitySectionSelection } from '../../../types/EntitySectionSelection';
 import type { Program, ProgramSession } from '../../../types/Program';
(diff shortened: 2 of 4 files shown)


web/components/TextBlock/Programs/Programs.module.css
```diff
@@ -1,9 +1,14 @@
 .venueNavContainer {
-  margin: 32px 0;
+  margin: 32px 0 48px;
+}
+
+.venueAddress {
+  text-align: center;
```
oyvind-stenhaug, line 6: Nit: It looks a bit odd to me to have this centered on mobile. Should this apply from `--medium-up` instead?

web/components/TextBlock/Programs/Programs.tsx
```diff
@@ -122,32 +122,45 @@ export const Programs = ({
...
         <div className={styles.venueNavContainer}>
-          <VenueNav
-            venues={venues}
-            activeVenue={activeProgram.venues[0]}
-            mainVenue={mainVenue}
-          />
+          <VenueNav {...{ activeVenue, venues, mainVenue }} />
         </div>
 
+        {typeof address === 'object' && (
```
oyvind-stenhaug, line 149: Is it ever not `object`? When nothing is filled in (e.g. London in current staging dataset), the query returns `null`, and the type of null is object.

Can it instead say

```suggestion
        {address && (
```

That way it should be unnecessary to have all the optional chaining/safe navigation below, and more importantly it avoids outputting empty elements when there is no address.

web/components/TextBlock/Programs/Programs.tsx
```diff
@@ -122,32 +122,45 @@ export const Programs = ({
...
+          <address className={styles.venueAddress}>
```
oyvind-stenhaug, line 150: This should not use `<address>`. That [represents contact information](https://html.spec.whatwg.org/multipage/sections.html#the-address-element) for (in this case) the entire document (e.g. for the /programs page).

I think it should be `<p>` or `<div>` instead. (If it's a `<p>` then it shouldn't contain divs, but see below)

web/components/TextBlock/Programs/Programs.tsx
```diff
@@ -122,32 +122,45 @@ export const Programs = ({
...
+            <div>{address?.name}</div>
```
oyvind-stenhaug, line 151: Addresses are actually one case where the [`<br>` element](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-br-element) is appropriate, according to the HTML spec.

It would be nice if these address parts could be joined by those instead of being output in separate divs, so we avoid empty divs in the cases where one part is missing. E.g. the missing `name` for Oslo in staging:

```
<div></div><div>Trondheimsveien 2k</div><div>Oslo, 0560</div>
```

vs

```
Trondheimsveien 2k<br>Oslo, 0560
```

Nit if it's hard to do.

web/types/Venue.ts
```diff
@@ -1,14 +1,39 @@
...
+import type { Section } from './Section';
+import type { Slug } from './Slug';
 
 export type Venue = {
+  _createdAt: string;
   _id: string;
+  _rev: string;
   _type: 'venue';
-  name: string;
+  _updatedAt: string;
```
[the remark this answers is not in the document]
j0nas: Yeah, this is a good point. I've been somewhat thoughtlessly inconsistent on this. Having an opportunity to reflect on this now, I think that the types should document all fields which are available. The alternative, if we ever needed to use the mentioned fields would be to create awkward ad-hoc union types. So your first suggestion makes the most sense to me. 8d33914

Thanks for pointing this out.

web/components/TextBlock/Programs/Programs.module.css
j0nas: f577bd8

web/components/TextBlock/Programs/Programs.tsx
j0nas: TIL! ffe6556

web/components/TextBlock/Programs/Programs.tsx
j0nas: No problem. 3ae7387

web/types/Venue.ts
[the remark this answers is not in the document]
oyvind-stenhaug: (Another alternative would be to add them as optional (`?`) if/when we ever need to use them and omit them entirely until then. We may not necessarily always want the web/ types to be 1:1 with the Sanity schema anyway, e.g. `coalesce`-ing things. But no objections or strong opinions in this case)

web/components/TextBlock/Programs/Programs.tsx
```diff
@@ -122,32 +123,53 @@ export const Programs = ({
...
 
+        {formattedAddress && (
+          <p className={styles.venueAddress}>
+            {formattedAddress.map((line, index) => (
+              <Fragment key={index}>
+                {line}
+                {index !== formattedAddress.length - 1 && <br />}
+              </Fragment>
+            ))}
+          </p>
```
oyvind-stenhaug, line 170: Ah, removing the left-alignment for narrow viewports uncovered the fact that this element didn't have any horizontal padding. Let's have a grid wrapper here?

```suggestion
          <GridWrapper>
            <p className={styles.venueAddress}>
              {formattedAddress.map((line, index) => (
                <Fragment key={index}>
                  {line}
                  {index !== formattedAddress.length - 1 && <br />}
                </Fragment>
              ))}
            </p>
          </GridWrapper>
```

oyvind-stenhaug (changes requested):

oyvind-stenhaug (approved):

